Issue #11, opened 2019-04-02
[BUG] GoTable - No Data Throws Error


mwagz:
### Problem
When setting up the `tableConfig`, if the data provided to the `tableData` key is or `undefined`, the TableComponent will throw an error.

```shell
GoTableComponent.html:5 ERROR TypeError: Cannot read property 'tableData' of undefined
    at Object.eval [as updateDirectives] (GoTableComponent.html:14)
    at Object.debugUpdateDirectives [as updateDirectives] (vendor.js:59923)
    at checkAndUpdateView (vendor.js:59381)
    at callViewAction (vendor.js:59604)
    at execComponentViewsAction (vendor.js:59576)
    at checkAndUpdateView (vendor.js:59381)
    at callViewAction (vendor.js:59604)
    at execEmbeddedViewsAction (vendor.js:59595)
    at checkAndUpdateView (vendor.js:59381)
    at callViewAction (vendor.js:59604)
```

Line 5 of the `GoTableComponent.html` file is:
```html
<tr *ngFor="let item of localTableConfig.tableData" class="go-table__tr”>
```

---

### Expected Behavior
The GoTableComponent should handle a "No Data" state.

---

### Reproduction Steps
**JavaScript**
```javascript
let myData;
let tableConfig = new GoTableConfig({
  tableData: myData
});
```

**HTML**
```html
<go-table [tableConfig]="tableConfig">
  <go-table-column field="label" title="Label"></go-table-column>
  <go-table-column field="another" title="Another"></go-table-column>
</go-table>
```

grahamhency:
Merged, closing.

PR #14, opened 2019-04-08
Added a 'no-data' state to the go-table

grahamhency:
Added a `noDataText` property to the `GoTableConfig` class to specify the text to show when there is no data.

For Issue #11

### DIFF
--- a/projects/go-lib/package.json
+++ b/projects/go-lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tangoe/goponents",
-  "version": "0.5.5",
+  "version": "0.5.6",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/mwagz/tangoe-shared.git"
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.ts
@@ -32,11 +32,10 @@ export class GoTableComponent implements OnInit, OnChanges {
   renderTable() {
     if (this.tableConfig) {
       this.localTableConfig = JSON.parse(JSON.stringify(this.tableConfig));
-      this.showTable = true;
       this.handleSort();
-    } else {
-      this.showTable = false;
     }
+    
+    this.showTable = Boolean(this.tableConfig);
   }
 
   hasData() : boolean {
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.ts
@@ -1,4 +1,4 @@
-import { Component, ContentChildren, Input, OnChanges, QueryList } from '@angular/core';
+import { Component, ContentChildren, Input, OnChanges, QueryList, OnInit } from '@angular/core';
 import { GoTableColumnComponent } from './go-table-column.component';
 import { GoTableConfig, SortDirection } from './index';
 import { sortBy } from './go-table-utils';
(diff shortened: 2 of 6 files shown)


projects/go-lib/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tangoe/goponents",
-  "version": "0.5.5",
+  "version": "0.5.6",
```
mwagz: Should this be ticked up for every addition? Or should it be ticked up on each release?

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -8,25 +8,45 @@ import { sortBy } from './go-table-utils';
...
   }
 
   renderTable() {
-    if (this.tableConfig!.tableData) {
+    if (this.tableConfig) {
       this.localTableConfig = JSON.parse(JSON.stringify(this.tableConfig));
+      this.showTable = true;
       this.handleSort();
+    } else {
+      this.showTable = false;
```
mwagz: Could we write this a little more concisely?

```javascript
if (this.tableConfig) {
  this.localTableConfig = JSON.parse(JSON.stringify(this.tableConfig));
  this.handleSort();
}

this.showTable = Boolean(this.tableConfig);
```

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -8,25 +8,45 @@ import { sortBy } from './go-table-utils';
...
     }
   }
 
+  hasData() : boolean {
+    if (this.localTableConfig) {
+      return this.localTableConfig.tableData ? this.localTableConfig.tableData.length > 0 : false;
```
mwagz: If `this.localTableConfig.tableData` is anything except 0, it will be true. `0` is falsey, so I _think_ we could write this like so:

```javascript
return this.localTableConfig.tableData.length;
```

Should we make sure `tableData` is present? We could use the TypeScript safe navigator...
```javascript
return this.localTableConfig?.tableData.length
```

projects/go-lib/package.json
grahamhency: I am considering this a hotfix, so I am incrementing the version so I can republish it. We do need to figure out the best way of how this should work though. 🤷‍♂️

projects/go-lib/src/lib/components/go-table/go-table.component.ts
grahamhency: This makes sense, updated.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
grahamhency: In this case, I am first checking `this.localTableConfig.tableData` to see if it exists. Then I get the length and evaluate further. This syntax:
`this.localTableConfig?.tableData.length`
 is invalid in TypeScript, there isn't a safe navigator that I'm aware of.

projects/go-lib/package.json
mwagz: It _seems_ like most open source projects will have a "Release PR" that contains this work. The [Angular Repo for example](https://github.com/sfeir-open-source/angular/pull/266/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R3).

projects/go-lib/src/lib/components/go-table/go-table.component.ts
mwagz: I had this wrong... I was referring to the [null assertion operator](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator) that you mentioned on March 25th in the Dojo.

```
this.localTableConfig!.tableData.length
```

I do see what you're doing, though, and without the null assertion thing, I mean that you don't need the ternary, really, because you can just check `length`. If `length` is any truthy value (0 is falsey), then it would return true, else it would return false. Below would be equivalent functionality, if you wanted to remove the ternary.

```typescript
if (this.localTableConfig && this.localTableConfig.tableData) {
  return this.localeTableConfig.tableData.length;
}

return false;
```

grahamhency: All good, I've updated this.

projects/go-lib/package.json
grahamhency: For the sake of this PR, we will keep this and push the new version asap after this is merged. We can figure out a better release structure moving forward.

PR #15, opened 2019-04-10
Create LICENSE

akfernun:
Create a license file

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2017 MOBI Wireless Management
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2017 MOBI Wireless Management
```
devchris: Should this be "Tangoe" instead?

akfernun: Umm... I think so. Honestly, I took this one from one of our other open source gems. https://github.com/mobi/telephone_number

akfernun:
@devchris Care to take another look? I guess I need to go back to our telephone_number gem and figure out what to do there as well.

PR #19, opened 2019-04-30
Table Pagination

grahamhency:
### This adds both client and server-side pagination capabilities.

As a result:
- Added server-side sorting capabilities
- Loading state

### To implement a simple client side paging:

```
// component.ts
let data = [{}] // array of things
let tableConfig = new GoTableConfig({
        tableData: data.results,
        totalCount: data.totalCount
      });
```
```
<!-- component.html -->
<go-table [tableConfig]="tableConfig">
```
This approach assumes you have all of the data already loaded into the component onInit.

### To implement simple server side paging:
```
// component.ts
tableConfig: GoTableConfig;
tableLoading: boolean = true;

ngOnInit() {
    this.appService.getMockData().subscribe(data => {
        dataMode: GoTableDataMode.server,
        tableData: data.results,
        totalCount: data.totalCount
      });

     this.tableLoading = false;
    })
  }

handleTableChange(currentTableConfig: GoTableConfig) : void {
  this.appService.getMockData(currentTableConfig).subscribe(data => {
    currentTableConfig.tableData = data.results;
    currentTableConfig.totalCount = data.totalCount;
      
    this.tableConfig = currentTableConfig;
    this.tableLoading = false;
  });
}
```
```
<!-- component.html -->
<go-table [loadingData]="tableLoading"
                  [tableConfig]="tableConfig"
                  (tableChange)="handleTableChange($event)"
                  *ngIf="tableConfig">
```
This approach expects server side pagination to exist. Here we're assuming that in the `service` for the component the correct mapping of `GoTableConfig`'s paging property to the endpoint's paging params will exist. Optionally, the endpoint's param names can be updated to match that of `GoTableConfig's`.


projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -46,23 +59,123 @@ export class GoTableComponent implements OnInit, OnChanges {
...
+
+    let beginning = this.localTableConfig.paging.skip + 1;
+    let endingEstimate = paging.skip + paging.perPage;
+    let ending = endingEstimate <= totalCount ? endingEstimate : totalCount - paging.skip;
+
+    return beginning + " - " + ending;
+  }
+
+  setDisplayData() : any[] {
+    let { dataMode, paging, tableData } = this.localTableConfig;
```
akfernun: yeah, I think some getters and setters might be nice here.

projects/go-lib/src/lib/components/go-table/go-table-config.ts
```diff
@@ -1,17 +1,33 @@
+import { GoTablePagingConfig } from './go-table-paging';
 import { GoTableSortConfig } from './go-table-sort';
 
 export class GoTableConfig {
+  dataMode?: GoTableDataMode = GoTableDataMode.client;
+  loadingData?: boolean = false;
   noDataText?: string = 'No Data';
+  pageable?: boolean = true;
+  paging?: GoTablePagingConfig = new GoTablePagingConfig();
```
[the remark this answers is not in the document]
grahamhency: Yup, absolutely. I've updated `sort` to `sortConfig` as well.

projects/go-lib/src/lib/components/go-table/go-table-config.ts
```diff
@@ -1,17 +1,33 @@
...
   sort?: GoTableSortConfig;
   sortable?: boolean = true;
   tableData: any[];
+  totalCount?: number;
 
   constructor(fields?: {
+    dataMode?: GoTableDataMode,
```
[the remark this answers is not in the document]
grahamhency: Yes, lol. First pass at naming variables usually sucks 🤷‍♂

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -46,23 +59,123 @@ export class GoTableComponent implements OnInit, OnChanges {
...
+      } else {
+        this.handleSort();
+        this.localTableConfig.loadingData = false;
+      }
+    }
+  }
+
+  nextPage() : void {
+    this.localTableConfig.loadingData = true;
+    this.localTableConfig.paging.skip = this.localTableConfig.paging.skip + this.localTableConfig.paging.perPage;
```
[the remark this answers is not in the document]
grahamhency: Yeah, that makes sense. I've changed it. Although this makes me think that we should change `perPage` to `limit` 🤷‍♂

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -46,23 +59,123 @@ export class GoTableComponent implements OnInit, OnChanges {
...
     return {
-      'go-table__th--sort-up': this.sortClasses(columnField, SortDirection.ascending),
-      'go-table__th--sort-down': this.sortClasses(columnField, SortDirection.descending)
+      'go-table__th--sort-down': this.sortClasses(columnField, SortDirection.ascending),
+      'go-table__th--sort-up': this.sortClasses(columnField, SortDirection.descending)
     }
   }
 
-  toggleSort(columnField: string) {
+  toggleSort(columnField: string) : void {
```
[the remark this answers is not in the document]
grahamhency: Yeah... personal preference to have a space before/after return type. 🤷‍♂ We can do it either way I suppose, just habit.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -46,23 +59,123 @@ export class GoTableComponent implements OnInit, OnChanges {
...
+    this.localTableConfig.paging.skip = 0;
+
+    if (this.isServerMode()) {
+      this.tableChange.emit(this.localTableConfig);
+    } else {
+      this.localTableConfig.loadingData = false;
+    }
+  }
+
+  setPerPage(e) : void {
```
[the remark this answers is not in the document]
grahamhency: Yep, switched it.

projects/go-lib/src/lib/components/go-table/go-table-paging.ts
```diff
@@ -0,0 +1,14 @@
+export class GoTablePagingConfig {
+  pageSizes: number[] = [20, 50, 100];
```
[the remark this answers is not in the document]
grahamhency: Mayyyybe, but it seems like we'd only use it in this one place?

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -12,6 +20,8 @@ export class GoTableComponent implements OnInit, OnChanges {
 
   @Input() tableConfig: GoTableConfig;
 
+  @Output() tableChange: EventEmitter<GoTableConfig> = new EventEmitter();
```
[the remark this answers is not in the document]
grahamhency: Yeah, oversight. I've updated it!

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -46,23 +59,123 @@ export class GoTableComponent implements OnInit, OnChanges {
...
+      this.localTableConfig.loadingData = false;
+    }
+  }
+
+  setPerPage(e) : void {
+    this.localTableConfig.loadingData = true;
+    this.localTableConfig.paging.perPage = Number(e.target.value);
+    this.localTableConfig.paging.skip = 0;
+
+    if (this.isServerMode()) {
```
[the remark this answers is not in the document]
grahamhency: Yes! Good catch, I was able to do this with most of the instances of it.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -85,4 +198,13 @@ export class GoTableComponent implements OnInit, OnChanges {
     return sort && sort.column === columnField && sort.direction === direction;
   }
 
+  private setTotalCount() : void {
+    let { totalCount, tableData } = this.localTableConfig; 
+
+    this.localTableConfig.totalCount = totalCount !== undefined && totalCount !== null ? totalCount : tableData.length;
```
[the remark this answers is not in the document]
grahamhency: Yeah, you're right because it's defaulted to `null` in the `GoTableConfig` class. Good catch there.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -46,23 +59,123 @@ export class GoTableComponent implements OnInit, OnChanges {
     return false;
   }
 
+  showPaging() : boolean {
+    return this.hasData() && this.localTableConfig.pageable;
```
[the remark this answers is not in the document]
grahamhency: Possibly. What benefit does that provide over the current implementation? Most of these are just seeing if the property is set.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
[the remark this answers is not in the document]
akfernun: Maybe...? 🤷‍♂

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -1,26 +1,51 @@
...
+    </table>
+  </div>
+  <div class="go-table-toolbar go-table-toolbar--bottom" *ngIf="showPaging()">
+    <div class="go-table-page-size">
+      <select class="go-form__select" (change)="setPerPage($event)">
+        <ng-container *ngFor="let size of localTableConfig.paging.pageSizes">
+          <option value="{{ size }}">{{ size }}</option>
+        </ng-container>
+      </select>
+      <label class="go-form__label go-form__label--table-toolbar">Page Size</label>
```
AlexOverbeck: I think this would be more appropriate as an element selector inside of the `go-table-toolbar`. I think an easy way to look at it is:
```
block == component of shared UI
element == chunk inside the component that isn't shared
modifier == alterations or additions to things that get applied when the state changes.
```

If we think about it like that, this `go-table-toolbar` could realistically be a block/component _(not that it should be it's own component)_. Most of these other things are elements inside of that block. It might end up looking something like this:
```html
<div class="go-table-toolbar" *ngIf="showPaging()">
  <div class="go-form go-table-toolbar__page-size">
    <select class="go-form__select" (change)="setPerPage($event)">
      <option [ngValue]="size" *ngFor="let size of localTableConfig.paging.pageSizes">
        {{ size }}
      </option>
    </select>
    <label class="go-form__label go-table-toolbar__label">Page Size</label>
  </div>
  <div class="go-table-toolbar__totals">
    {{ outputResultsPerPage() }} /
    {{ localTableConfig.totalCount | number : '1.0-0' }} Results
  </div>
  <div class="go-table-page-controls">
    <go-icon class="go-table-page-controls__icon" [ngClass]="{ 'go-table-page-controls__icon--disabled': isFirstPage() }" (click)="setFirstPage()" icon="first_page"></go-icon>
    <go-icon class="go-table-page-controls__icon" [ngClass]="{ 'go-table-page-controls__icon--disabled': isFirstPage() }" (click)="prevPage()" icon="chevron_left"></go-icon>
    <go-icon class="go-table-page-controls__icon" [ngClass]="{ 'go-table-page-controls__icon--disabled': isLastPage() }" (click)="nextPage()" icon="chevron_right"></go-icon>
    <go-icon class="go-table-page-controls__icon" [ngClass]="{ 'go-table-page-controls__icon--disabled': isLastPage() }" (click)="setLastPage()" icon="last_page"></go-icon>
  </div>
</div>
```

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -1,26 +1,51 @@
...
+            </ng-template>
+          </td>
+        </tr>
+      </tbody>
+    </table>
+  </div>
+  <div class="go-table-toolbar go-table-toolbar--bottom" *ngIf="showPaging()">
+    <div class="go-table-page-size">
+      <select class="go-form__select" (change)="setPerPage($event)">
+        <ng-container *ngFor="let size of localTableConfig.paging.pageSizes">
```
AlexOverbeck: What is the purpose of this `ng-container`? Couldn't this be written:
```html
<option [ngValue]="size" *ngFor="let size of localTableConfig.paging.pageSizes">{{size}}</option>
```
Additionally, I think if we use `ngValue` it prevents a potential case where the text "{{ size }}" could be evaluated as the value of this input.

projects/go-lib/src/lib/components/go-table/go-table.component.scss
```diff
@@ -45,24 +60,85 @@
...
+  @include transition(all);
+
+  &:hover {
+    background: $theme-light-bg-hover;
   }
+}
 
-  &:last-child {
-    border-right: 1px solid $theme-light-border;  
+.go-table-page-controls__icon--disabled {
```
AlexOverbeck: This will be able to be refactored out with the updates we are making to the go-icon.

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -1,26 +1,51 @@
...
-          </ng-template>
-        </td>
-      </tr>
-    </tbody>
-  </table>
+  <div class="go-table__table">
+    <table class="go-table" *ngIf="hasData(); else noDataTable">
+      <thead>
+        <tr>
+          <th *ngFor="let col of columns" class="go-table__th" [ngClass]="columnClasses(col.field)" (click)="toggleSort(col.field)">
```
AlexOverbeck: What if we called this class `go-table__heading`?

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -1,26 +1,51 @@
...
+            <ng-container *ngIf="!col.goTableHead; else columnHeaderOutlet">{{ col.title || col.field }}</ng-container>
+            <ng-template #columnHeaderOutlet>
+              <ng-container *ngTemplateOutlet="col.goTableHead"></ng-container>
+            </ng-template>
+          </th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr *ngFor="let item of setDisplayData()" class="go-table__tr">
+          <td *ngFor="let col of columns" class="go-table__td" width="{{ col.width || 'auto' }}">
```
AlexOverbeck: What if we called this class `go-table__column`?

projects/go-lib/src/lib/components/go-table/go-table.component.html
AlexOverbeck: In the past I've used a simple svg to establish the same effect as the `[ngClass]="columnClasses(col.field)"`. We could potentially even wrap that svg in a button to get bonus accessibility points. I don't know that we need to make this change now, but it could alternatively look something like this:
```html
<svg height="20" viewBox="0 0 24 24">
  <desc>up and down arrows for controlling table column sort direction</desc>
  <polygon points='9 3 5 6.99 8 6.99 8 14 10 14 10 6.99 13 6.99 9 3'></polygon>
  <polygon points='15 21 19 17.01 16 17.01 16 10 14 10 14 17.01 11 17.01 15 21'></polygon>
</svg>
```
The benefit being more control over the state of the UI and some additional accessibility.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -11,6 +19,9 @@ import { sortBy } from './go-table-utils';
 export class GoTableComponent implements OnInit, OnChanges {
 
   @Input() tableConfig: GoTableConfig;
+  @Input() loadingData: boolean = false;
```
akfernun: Why are we allowing this to be an input?

projects/go-lib/src/lib/components/go-table/go-table.component.html
grahamhency: - I thought about `go-table__heading`, but I felt that we _could_ possibly have a heading that's not a `th` in which case, we could reserve that class for such a case. I also thought that `th` was very specific. These styles should only be applied to a `th`.

- This is a pretty good idea I think we can make use of in the redesign of the table.

projects/go-lib/src/lib/components/go-table/go-table.component.html
grahamhency: Thought about this, kind of along the same lines as my comment about the `heading`

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -1,26 +1,51 @@
...
+            <ng-container *ngIf="!col.goTableCell; else columnCellOutlet">{{ col.getFieldData(item) }}</ng-container>
+            <ng-template #columnCellOutlet>
+              <ng-container *ngTemplateOutlet="col.goTableCell; context: { $implicit: item }"></ng-container>
+            </ng-template>
+          </td>
+        </tr>
+      </tbody>
+    </table>
+  </div>
+  <div class="go-table-toolbar go-table-toolbar--bottom" *ngIf="showPaging()">
```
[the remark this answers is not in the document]
grahamhency: The thought here is that the toolbar should be able to stand on it's own and that we _could_ have a top toolbar as well where the styles would need to be opposite that of `go-table-toolbar--bottom`.

projects/go-lib/src/lib/components/go-table/go-table.component.html
grahamhency: Yeah, that does make sense. I've change it to `go-table-toolbar__label`.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
grahamhency: @akfernun so that we can control the loading state dependent upon server requests.

projects/go-lib/src/lib/components/go-table/go-table-paging.ts
```diff
@@ -0,0 +1,14 @@
+export class GoTablePagingConfig {
```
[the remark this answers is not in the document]
grahamhency: Yes, you're totally right here. Originally I was exploring options between interfaces and classes for this implementation and I ended up classes, but missed updating the naming convention on the file completely.

projects/go-lib/src/lib/components/go-table/go-table.component.html
grahamhency: Yep, you're right here too. I've removed the `ng-container`. :+1:

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -46,23 +59,123 @@ export class GoTableComponent implements OnInit, OnChanges {
...
+      this.tableChange.emit(this.localTableConfig);
+    } else {
+      this.localTableConfig.loadingData = false;
+    }
+  }
+
+  outputResultsPerPage() : string {
+    let { paging, totalCount } = this.localTableConfig;
+
+    let beginning = this.localTableConfig.paging.skip + 1;
```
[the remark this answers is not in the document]
grahamhency: Yes, I've updated them.

projects/go-lib/src/lib/components/go-table/go-table.component.html
AlexOverbeck: Looks like `ngValue` isn't quite right here. We could update this to use `[value]="size"` though if we wanted to avoid the above mentioned edge case.

projects/go-lib/src/lib/components/go-table/go-table.component.html
AlexOverbeck: Seems like a valid assumption, but maybe one that we don't need to make yet?

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -1,26 +1,50 @@
...
+    <div class="go-table-page-size">
```
AlexOverbeck: This feels like an element of the `go-table-toolbar`. would this make more sense as `go-table-toolbar__page-size`?

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -1,26 +1,50 @@
...
+      <select class="go-form__select" (change)="setPerPage($event)">
+        <option *ngFor="let size of localTableConfig.pageConfig.pageSizes"
+                value="{{ size }}">{{ size }}</option>
+      </select>
+      <label class="go-form__label go-table-toolbar__label">Page Size</label>
+    </div>
+    <div class="go-table-totals">
```
AlexOverbeck: This feels like an element of the `go-table-toolbar`. would this make more sense as `go-table-toolbar__totals`?

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -1,26 +1,50 @@
...
+      {{ outputResultsPerPage() }} /
+      {{ localTableConfig.totalCount | number : '1.0-0' }} Results
+    </div>
+    <div class="go-table-page-controls">
```
AlexOverbeck: This feels like an element of the `go-table-toolbar`. would this make more sense as `go-table-toolbar__controls`?

projects/go-tester/src/styles.scss
```diff
@@ -2,6 +2,8 @@
 @import url('https://fonts.googleapis.com/css?family=Roboto:300,400,700');
 @import url('https://fonts.googleapis.com/icon?family=Material+Icons');
 
+@import "~@tangoe/gosheets/gosheets";
```
AlexOverbeck: Nice! This should allow us to refactor out some additional code.

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -1,26 +1,50 @@
...
+      <go-icon class="go-table-page-controls__icon" [ngClass]="{ 'go-table-page-controls__icon--disabled': isFirstPage() }" (click)="setFirstPage()" icon="first_page"></go-icon>
+      <go-icon class="go-table-page-controls__icon" [ngClass]="{ 'go-table-page-controls__icon--disabled': isFirstPage() }" (click)="prevPage()" icon="chevron_left"></go-icon>
```
AlexOverbeck: These would be better if they were wrapped in a button. This would give you access to these controls via the keyboard.

projects/go-lib/src/lib/components/go-table/go-table.component.html
grahamhency: Yes! Updated it. :+1:

projects/go-lib/src/lib/components/go-table/go-table.component.html
grahamhency: Yes! Updated it. :+1:

projects/go-lib/src/lib/components/go-table/go-table.component.html
grahamhency: Yes! Updated it. :+1:

projects/go-tester/src/styles.scss
grahamhency: Yessssssir!

projects/go-lib/src/lib/components/go-table/go-table.component.html
grahamhency: Great idea! Wish I would've thought about that first. I've added buttons here and it's a much better UX. 🥇  :+1:

PR #21, opened 2019-05-06
Feature toasts

grahamhency:
This PR adds the `go-toast` component and corresponding `go-toaster` service.

To use a toast without the toaster:
```
<go-toast header="Hello There!"
                  message="Did you know that Obi-Wan Kenobi was super cool?">
</go-toast>
```

There are 3 available `type`s:
- 'neutral' _(default)_
- 'positive' _to be used for successful things_
- 'negative' _to be used fore errors/warnings/etc_

There is also a `dismissable` binding and `(handleDismiss)` event. To use this, set `dismissable` to `true` and capture the `handleDismiss` event in your component.

To use the go-toaster service:
```
// component.ts
import { GoToasterService } from '@tangoe/goponents`;

constructor(private goToasterService: GoToasterService) { }

ngOnInit() {
  this.goToasterService.toastSuccess({ message: 'It works!' });
  this.goToasterService.toastInfo({ message: 'Just so you know...' });
  this.goToasterService.toastError({ message: 'This broke everything' });
}
```


projects/go-lib/src/lib/components/go-toast/go-toast.component.html
```diff
@@ -0,0 +1,13 @@
...
+  <div class="go-toast-strip" [ngClass]="statusClasses()"></div>
+  <div class="go-toast-status">
+    <go-icon class="go-toast-status__icon" [ngClass]="statusClasses()" icon="{{ goIconType() }}"></go-icon>
+  </div>
+  <div class="go-toast-content"  [ngClass]="{ 'go-toast-content--no-title': !header }">
+    <h5 class="go-toast-content__title" *ngIf="header">{{ header }}</h5>
+    <p class="go-toast-content__message">{{ message }}</p>
+  </div>
+  <div class="go-toast-dismiss" *ngIf="dismissable">
+    <go-icon class="go-toast-dismiss__icon" icon="close" (click)="dismiss()"></go-icon>
```
AlexOverbeck: This should be wrapped in a button so we know that it is clickable. We will also get access to this action via that keyboard.

projects/go-lib/src/lib/components/go-toast/go-toast.component.scss
```diff
@@ -0,0 +1,95 @@
...
+}
+
+.go-toast-dismiss {
+  padding: 0.25rem;
+  position: absolute;
+  right: 0;
+  top: 0;
+}
+
+.go-toast-dismiss__icon {
```
AlexOverbeck: These styles feel like they could probably be added to some sort of button block level thing. We can do that in a separate PR, but it seems like this could be something like:
```html
<button class="go-icon-button">
  <i class="go-icon go-icon-button__icon">icon</i>
</button>
```

projects/go-lib/src/lib/components/go-toaster/go-toaster.service.ts
```diff
@@ -0,0 +1,79 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class GoToasterService {
+
+  toasts: GoToastComponent[] = [];
+
+  constructor() { }
+
+  /**
```
AlexOverbeck: I'm liking all of these comments. We should start doing this everywhere.

projects/go-lib/src/lib/components/go-toaster/go-toaster.service.ts
```diff
@@ -0,0 +1,79 @@
...
+  private setupTimer(comp: GoToastComponent, showPeriod: number) :void {
+    timer(showPeriod).subscribe(() => {
+      this.toasts.splice(this.toasts.indexOf(comp), 1);
+    });
+  }
+
+  //#endregion
+}
+
+interface ToastInterface {
```
AlexOverbeck: Should this live in the same file or should it be separated out? I don't yet know enough one way or the other.

projects/go-lib/src/lib/components/go-toast/go-toast.component.html
```diff
@@ -0,0 +1,13 @@
+<div class="go-toast">
+  <div class="go-toast-strip" [ngClass]="statusClasses()"></div>
```
AlexOverbeck: We could probably sprinkle in some more `.block__element`s in this file, but this makes sense to me.

projects/go-lib/src/lib/components/go-toaster/go-toaster.model.ts
```diff
@@ -0,0 +1,5 @@
+export interface ToastInterface {
+  header?: string;
+  message?: string;
+  type?: string;
+}
```
AlexOverbeck: Should any of these be required?

projects/go-lib/src/lib/services/go-message/go-message.module.ts
```diff
@@ -5,7 +5,9 @@ import { GoMessageService } from './go-message.service';
 
 @NgModule({
   declarations: [],
-  providers: [GoMessageService],
+  providers: [
+    GoMessageService
```
AlexOverbeck: Do we still need this if the `GoMessageService` is `providedIn: 'root'`?

projects/go-tester/src/styles.scss
```diff
@@ -1,9 +1,11 @@
 /* You can add global styles to this file, and also import other style files */
 @import url('https://fonts.googleapis.com/css?family=Roboto:300,400,700');
 @import url('https://fonts.googleapis.com/icon?family=Material+Icons');
+@import "~@tangoe/gosheets/gosheets";
 
 @import "~@tangoe/gosheets/gosheets";
```
AlexOverbeck: Duplicate imports.

C-658VsXoAo3ovC

projects/go-lib/src/lib/components/go-toaster/go-toaster.model.ts
grahamhency: Thought about that, this interface is only used in the service where the type and message are always set. Because those methods take a parameter of type ToastInterface it would require the user to pass these when calling the method.

projects/go-lib/src/lib/services/go-message/go-message.module.ts
grahamhency: This is actually going to be deprecated by the ToasterService so I'm not too worried about it.

projects/go-tester/src/styles.scss
grahamhency: 🤦‍♂ good catch and nice gif choice!

PR #26, opened 2019-05-09
Replaces icon and buttons styles with gosheets classes

AlexOverbeck:
https://mobi.teamwork.com/#tasks/12980128

This refactors the buttons and icon components to accept the BEM
style classes defined in the gosheets module. It also fixes a bug
with the button component where the buttonVariant was being applied
to the buttons type attribute.


projects/go-lib/src/lib/components/go-button/go-button.component.ts
```diff
@@ -18,22 +17,28 @@ export class GoButtonComponent {
 
   constructor() {}
 
-  public clicked() : void {
-    if (this.isProcessing || this.buttonDisabled) { return; }
```
AlexOverbeck: We don't actually need this protection if we utilize the disabled attribute properly.

projects/go-lib/src/lib/components/go-icon/go-icon.component.ts
```diff
@@ -1,10 +1,25 @@
...
-  templateUrl: 'go-icon.component.html',
-  styleUrls: ['go-icon.component.scss']
+  templateUrl: 'go-icon.component.html'
 })
 export class GoIconComponent {
   @Input() icon: string;
+  @Input() iconModifier: string;
+  @Input() iconClass: string;
+
+  public classObject(): object {
```
AlexOverbeck: We should probably think of a consistent, potentially more performant way to handle this sort of class manipulation.

projects/go-lib/src/lib/components/go-button/go-button.component.ts
```diff
@@ -18,22 +17,28 @@ export class GoButtonComponent {
...
   
-  public reset() : void {
+  public reset(): void {
     this.isProcessing = false;
   }
 
-  public classList() : any {
+  public classObject(): object {
+    // 'alert' as a variant is depreciated and
+    // will be removed in a later version
```
grahamhency: I like this concept, allows us to communicate deprecated changes ahead of time. :+1:

projects/go-lib/src/lib/components/go-icon/go-icon.component.html
```diff
@@ -1 +1,6 @@
-<i class="material-icons">{{icon}}</i>
+<i
+  class="go-icon"
+  [ngClass]="classObject()"
+>
```
grahamhency: Still not sure how I feel about this particular syntax. 🤷‍♂

projects/go-lib/src/lib/components/go-icon/go-icon.component.ts
```diff
@@ -1,10 +1,25 @@
...
 @Component({
   selector: 'go-icon',
-  templateUrl: 'go-icon.component.html',
-  styleUrls: ['go-icon.component.scss']
+  templateUrl: 'go-icon.component.html'
 })
 export class GoIconComponent {
   @Input() icon: string;
+  @Input() iconModifier: string;
+  @Input() iconClass: string;
```
grahamhency: Should we alphabetize these `@Input`s?

projects/go-lib/src/lib/components/go-icon/go-icon.component.ts
grahamhency: Yeah, I don't really love this implementation... but understand the reasoning for it and at the current point in time I can't come up with a better alternative.

projects/go-lib/src/lib/components/go-icon/go-icon.component.ts
```diff
@@ -1,10 +1,25 @@
-import { Component, Input } from '@angular/core';
+import { Component, Input, OnInit } from '@angular/core';
 
 @Component({
   selector: 'go-icon',
-  templateUrl: 'go-icon.component.html',
-  styleUrls: ['go-icon.component.scss']
```
grahamhency: We need the contents of this sass file. It allows us to properly override the font-size of the icon

grahamhency: We adjusted the styling in `gosheets` to fix the issue we were overriding in this file, thus we now do not need it.

PR #43, opened 2019-06-05
Create Go Loader Component

grahamhency:
Built with SVG
Applied proper animation with updated loader to table changes
Organized go-tester page


projects/go-lib/src/lib/components/go-loader/go-loader.component.html
```diff
@@ -0,0 +1,34 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+     xmlns:xlink="http://www.w3.org/1999/xlink"
+     class="go-loader-container"
+     [ngClass]="containerClasses()"
+     viewBox="0 0 150 150">
+  <title>Loading</title>
+  <linearGradient id="goLoaderNeutralGradient">
+    <stop offset="15%" stop-color="#8A4EDE" />
```
StevenUlmer: I wonder if it's possible to define these colors in our css so we can use our scss variables? Something like
```
#goLoaderNeutralGradient {
 stop-color: $ui-color-positive;
}
```

grahamhency: @StevenUlmer good idea, I was able to create classes for the `stop-color`. I was not able to add the `offset` attribute to css though.

PR #47, opened 2019-06-25
Removes the ngx-toaster module and all references to it in goponents

AlexOverbeck:
We have successfully replaced our ngx-toaster dependency with our
own go-toaster service. We no longer need ngx-toaster.


grahamhency:
@AlexOverbeck the build fails here because it's trying to import GoMessage service in the `public_api.ts` file. Just found this out locally:

```
// Services
export * from './lib/services/go-message/go-message.service';
export * from './lib/services/go-message/go-message.module';
```

AlexOverbeck:
@grahamhency should be all fixed in #49

PR #55, opened 2019-07-09
[Feature] Side Navigation

StevenUlmer:
This implements a side nav that can be dropped into a layout.

To use it you would drop the side navigation into the app template as the first thing in the content.

```
<go-side-nav [menuItems]="menuItems"></go-side-nav>
```

It takes one parameter which is menuItems which takes a list of things that goes into the side nav. These can either be a group of items or an item. A group would have an array of subRoutes while a normal item would have a route and a routeTitle. The route is the url that the button goes to while the routeTitle is what is displayed on the button. Each item or group on the top level of the navigation needs to have a routeIcon attached to it for the purposes of collapsing the navigation. If a group doesn't have any subroutes it will be hidden by default.

ex list of menuItems

```
[
  { route: 'getting-started', routeIcon: 'power_settings_new', routeTitle: 'Getting Started' },
  { routeIcon: 'gavel', routeTitle: 'Standards', subRoutes: [
    { route: 'standards/colors', routeTitle: 'Colors' },
  ]}
]
```

### DIFF
--- a/projects/go-lib/src/lib/components/go-side-nav/go-nav-group/go-nav-group.component.scss
+++ b/projects/go-lib/src/lib/components/go-side-nav/go-nav-group/go-nav-group.component.scss
@@ -1,0 +1,60 @@
+@import '~@tangoe/gosheets/base/variables';
+@import '~@tangoe/gosheets/base/mixins';
+@import '../../../../styles/variables';
+
+.go-nav-group__dropdown {
+  align-items: center;
+  display: flex;
+  padding-right: .5rem;
+  user-select: none;
+  @include transition(all);
+
+  &:hover {
+    background: $theme-dark-bg-hover;
+  }
+}
+
+.go-nav-group__link {
+  display: flex;
+  flex-grow: 1;
+}
+
+.go-nav-group__title {
+  font-weight: $weight-light;
+  letter-spacing: $side-nav-letter-spacing;
+  padding: $outer-side-nav-padding;
+}
+
+.go-nav-group__icon {
+  font-size: 1.2rem;
+  padding: 1rem;
+}
+
+.go-nav-group__expand-icon {
+  border-radius: 50%;
+  color: $theme-dark-color;
+  cursor: pointer;
+  font-size: 1.5rem;
+  height: 2.5rem;
+  padding: 0.5rem;
+  width: 2.5rem;
+
+  &:hover {
+    background: $theme-dark-bg;
+  }
+}
+
+.go-nav-group__expand-icon--expanded {
+  transform: rotate(180deg);
+}
+
+.go-nav-group__inner-group {
+  .go-nav-group__title {
+    font-size: $inner-side-nav-font-size;
+    padding: $inner-side-nav-padding;
+  }
+}
+
+.go-nav-group--expanded {
+  background: $theme-dark-bg-active;
+}
--- a/projects/go-lib/src/lib/components/go-side-nav/go-nav-group/go-nav-group.component.ts
+++ b/projects/go-lib/src/lib/components/go-side-nav/go-nav-group/go-nav-group.component.ts
@@ -1,0 +1,30 @@
+import { Component, Input, ViewEncapsulation, Output, EventEmitter } from '@angular/core';
+import { NavGroup } from '../nav-group.model';
+import { NavItem } from '../nav-item.model';
+import { GoSideNavService } from '../go-side-nav/go-side-nav.service';
+
+@Component({
+  selector: 'go-nav-group',
+  templateUrl: './go-nav-group.component.html',
+  styleUrls: ['./go-nav-group.component.scss'],
+  encapsulation: ViewEncapsulation.None
+})
+export class GoNavGroupComponent {
+  @Input() navItem: NavGroup | NavItem;
+  @Input() class: string;
+  @Output() closeNavs = new EventEmitter<NavGroup>();
+
+  constructor (public navService: GoSideNavService) { }
+
+  expand(navGroup: NavGroup): void {
+    navGroup.expanded = !navGroup.expanded;
+
+    if (!this.navService.navOpen) {
+      this.navService.toggleNav();
+    }
+
+    if (navGroup.expanded) {
+      this.closeNavs.emit(navGroup);
+    }
+  }
+}
--- a/projects/go-lib/src/lib/components/go-side-nav/go-nav-item/go-nav-item.component.scss
+++ b/projects/go-lib/src/lib/components/go-side-nav/go-nav-item/go-nav-item.component.scss
@@ -1,0 +1,65 @@
+@import '~@tangoe/gosheets/base/variables';
+@import '~@tangoe/gosheets/base/mixins';
+@import '../../../../styles/variables';
+
+.go-nav-item {
+  align-items: center;
+  display: flex;
+  padding-right: .5rem;
+  @include transition(all);
+
+  &:hover {
+    background: $theme-dark-bg-hover;
+  }
+}
+
+.go-nav-item__link::before {
+  background: $base-primary;
+  border-radius: 0 $global-radius $global-radius 0;
+  content: " ";
+  height: 100%;
+  left: 0;
+  opacity: 0;
+  position: absolute;
+  top: 0;
+  width: 4px;
+  @include transition(all);
+}
+
+.go-nav-item__link--active {
+  .go-nav-item__title {
+    font-weight: $weight-regular;
+  }
+}
+
+.go-nav-item__link--active::before {
+  opacity: 1;
+}
+
+.go-nav-item__link,
+.go-nav-item__link:visited,
+.go-nav-item__link:focus,
+.go-nav-item__link:active {
+  color: $theme-dark-color;
+}
+
+.go-nav-item__link {
+  align-items: center;
+  border: none;
+  display: flex;
+  flex-grow: 1;
+  position: relative;
+  text-decoration: none;
+}
+
+.go-nav-item__title {
+  font-size: $inner-side-nav-font-size;
+  font-weight: $weight-light;
+  letter-spacing: $side-nav-letter-spacing;
+  padding: $inner-side-nav-padding;
+}
+
+.go-nav-item__title--with-icon {
+  font-size: 1rem;
+  padding: $outer-side-nav-padding;
+}
--- a/projects/go-lib/src/lib/components/go-side-nav/go-side-nav/go-side-nav.component.ts
+++ b/projects/go-lib/src/lib/components/go-side-nav/go-side-nav/go-side-nav.component.ts
@@ -1,0 +1,77 @@
+import { Component, OnInit, Input } from '@angular/core';
+import { Router, NavigationEnd } from '@angular/router';
+import { filter } from 'rxjs/operators';
+import { NavGroup } from '../nav-group.model';
+import { NavItem } from '../nav-item.model';
+import { GoSideNavService } from './go-side-nav.service';
+
+@Component({
+  selector: 'go-side-nav',
+  templateUrl: './go-side-nav.component.html',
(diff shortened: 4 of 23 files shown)


projects/go-lib/src/lib/components/go-side-nav/go-side-nav/go-side-nav.component.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  closeNavs(item) {
+    this.menuItems.forEach(group => {
+      (group as NavGroup).expanded = this.openAccordion((group as NavGroup), item);
+    });
+  }
+
+  //#region Private Methods
+
+  private setExpanded(route: NavGroup | NavItem, url: string) {
```
devchris: Can we name 'route' something different in this file? `route` is also something built-in in angular and this might be confusing for someone else in the future

projects/go-lib/src/lib/components/go-side-nav/go-side-nav/go-side-nav.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  private navGroupExpansion(route: NavGroup, url: string) {
+    route.expanded = route.subRoutes.some(subRoute => {
+      return this.setExpanded(subRoute, url);
+    });
+    return route.expanded;
+  }
+
+  private openAccordion(group: NavGroup, item: NavGroup): boolean {
```
devchris: This whole function seems to be very complex. Should we leave a comment above or somewhere in our documentation what this actually does?

I tried to come up with a better solution but couldn't make it dynamically. Maybe @grahamhency has another suggestion for this?

projects/go-lib/src/lib/components/go-side-nav/go-side-nav/go-side-nav.service.ts
```diff
@@ -0,0 +1,17 @@
...
+
+@Injectable({
+  providedIn: 'root'
+})
+
+export class GoSideNavService {
+  public navState = true;
+
+  constructor(
+  ) { }
```
devchris: Can we move this on one line only if our constructor is empty?

projects/go-lib/src/lib/components/go-side-nav/nav-item.model.ts
```diff
@@ -0,0 +1,6 @@
+export interface NavItem {
+  route: string;
+  routeTitle: string;
+  routeIcon?: string;
+  description?: string;
```
devchris: Should we alphabetize interfaces?

projects/go-tester/src/app/app.component.ts
```diff
@@ -26,10 +29,38 @@ export class AppComponent implements OnInit {
   tableConfig: GoTableConfig;
   tableLoading: boolean = true;
 
+  menuItems: Array<NavGroup | NavItem> = [
```
devchris: I don't think this is the right place to add our `menuItems`. Should we add somewhere else instead?

projects/go-tester/src/app/app.component.ts
```diff
@@ -81,4 +112,8 @@ export class AppComponent implements OnInit {
       }
     });
   }
+
+  toggleSideMenu(): void {
```
devchris: Should we also move this code outside of the `app.component`?

projects/go-lib/src/lib/components/go-side-nav/go-nav-group/go-nav-group.component.scss
```diff
@@ -0,0 +1,59 @@
...
+
+.go-nav-group--inner-group {
+  .go-nav-group__title {
+    letter-spacing: .02rem;
+    padding: .6rem .833rem .6rem 3.2rem;
+    font-size: 0.833rem;
+  }
+}
+
+.go-nav-group--expanded {
```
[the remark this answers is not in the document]
StevenUlmer: I think this is legit BEM as we are modifying the go-nav-group element

projects/go-lib/src/lib/components/go-side-nav/go-side-nav/go-side-nav.component.ts
```diff
@@ -0,0 +1,76 @@
...
+   * closes the other accordions that were previously open if they are not above the group that was
+   * clicked on. It uses recursion to go through nested groups to make sure that if a child group
+   * is open the parents of the child is also open.
+   * @param group this is the group we are trying to decide if it should be open.
+   * @param item this is the group that we are searching for that was clicked on and needs opened.
+   */
+
+  private openAccordion(group: NavGroup, item: NavGroup): boolean {
+    if (group.subRoutes) {
+      group.expanded = group.routeTitle === item.routeTitle;
```
grahamhency: I don't think we need this line. The `if/else` block below does the same thing.

projects/go-lib/src/lib/components/go-side-nav/go-side-nav/go-side-nav.component.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (group.routeTitle !== item.routeTitle) {
+        group.expanded = group.subRoutes.some(subRoute => {
+          return this.openAccordion((subRoute as NavGroup), item);
```
grahamhency: The only other way I could think to _maybe_ do this would be to use `EventEmitter` and work up/down the chain. :shrug: This is probably okay for now.

projects/go-lib/src/lib/components/go-side-nav/go-side-nav/go-side-nav.service.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from '@angular/core';
+import { Observable, of as observableOf, BehaviorSubject } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root'
+})
+
+export class GoSideNavService {
+  public navState = true;
```
grahamhency: Do you think we should change this to be a bit more specific as to what this is? Maybe `navOpen`?

projects/go-lib/src/public_api.ts
```diff
@@ -36,6 +36,12 @@ export * from './lib/components/go-off-canvas/go-off-canvas.component';
 export * from './lib/components/go-off-canvas/go-off-canvas.module';
 export * from './lib/components/go-off-canvas/go-off-canvas.service';
 
+// Side Nav
+export * from './lib/components/go-side-nav/go-side-nav.module';
```
grahamhency: These are good, might want to export `go-side-nav.component` as well.

projects/go-tester/src/app/app.component.ts
grahamhency: @devchris this is 💯 for testing, this `go-tester` project will never go to production anywhere. This is our playground so I think it's fine.

projects/go-tester/src/app/app.component.ts
grahamhency: See above comment @devchris, but again, this is just for testing.

projects/go-lib/src/public_api.ts
StevenUlmer: @grahamhency is there a reason to export the component outside of the module?

projects/go-lib/src/lib/components/go-side-nav/go-nav-item/go-nav-item.component.scss
```diff
@@ -0,0 +1,54 @@
...
+  content: " ";
+  height: 100%;
+  left: 0;
+  opacity: 0;
+  position: absolute;
+  top: 0;
+  width: 4px;
+}
+
+.go-nav-item__link--active::before {
```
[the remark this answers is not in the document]
StevenUlmer: Changing this to 
```
go-nav-item__link::before {
  &.go-nav-item__link--active {
    opacity: 1;
  }
}
```
actually doesn't work. The active icon to the left doesn't show up anymore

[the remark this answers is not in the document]
grahamhency: @StevenUlmer yeah, my bad. It would need to be:
```
.go-nav-item__link--active:before {
  opacity: 1;
}
```
You should only need to style the active modifier, just make sure it is below the declaration of `go-nav-item__link` in the scss file.

projects/go-lib/src/public_api.ts
grahamhency: We talked about it offline, but for the sake of documentation...

I could see `GoSideNavComponent` being utilized within another component as maybe a `ViewChild` or `ContentChildren` and to be able to type that correctly we would need access to that component. With a package like this, the best way to expose classes is through the `public_api.ts` file. This allows us to import like:
`import { GoSideNavComponent } from '@tangoe/goponents';`

projects/go-lib/src/lib/components/go-side-nav/go-nav-item/go-nav-item.component.scss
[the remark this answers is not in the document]
AlexOverbeck:
```css
.go-nav-item__link--active::before {
  opacity: 1;
}
```
`:` for pseudo classes. `::` for pseudo elements

projects/go-lib/src/lib/components/go-side-nav/go-nav-item/go-nav-item.component.scss
```diff
@@ -0,0 +1,54 @@
...
+}
+
+.go-nav-item__title {
+  font-weight: lighter;
+  letter-spacing: .02rem;
+  padding: .6rem .833rem .6rem 3.2rem;
+  font-size: 0.833rem;
+}
+
+.go-nav-item__title--with-icon {
```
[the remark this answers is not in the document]
AlexOverbeck: This one is a little odd, but I think it makes sense to me. the `go-nav-item` block has a `title` that needs to be a little different when it is `with-icon`.

projects/go-lib/src/lib/components/go-side-nav/_nav-variables.scss
```diff
@@ -0,0 +1,4 @@
+$inner-font-size: 0.833rem;
```
AlexOverbeck: I don't know how I feel about all of these variables being in a separate file.
1. Are there any variables in gosheets we could use for this?
2. We should probably pull these out into a more global settings file.

projects/go-lib/src/lib/components/go-side-nav/go-nav-group/go-nav-group.component.ts
```diff
@@ -0,0 +1,32 @@
...
+import { NavItem } from '../nav-item.model';
+import { GoSideNavService } from '../go-side-nav/go-side-nav.service';
+
+@Component({
+  selector: 'go-nav-group',
+  templateUrl: './go-nav-group.component.html',
+  styleUrls: ['./go-nav-group.component.scss'],
+  encapsulation: ViewEncapsulation.None
+})
+export class GoNavGroupComponent {
```
AlexOverbeck: There are a few things in this class that don't have typing. If we want to get the most out of TypeScript we should apply types wherever we can.
```typescript
// is this variable used anywhere?
expanded: boolean = false;

expand(navGroup: NavGroup): void {
```

StevenUlmer:
@grahamhency we talked about this in the office but for documentation sake

> What happens if someone provides an empty array for a NavGroup's subRoutes?

I think if someone provides an empty array of subRoutes we should just hide the accordion. This will make it easier to implement where someone doesn't have to worry about removing a top level object if all of the subroutes are supposed to be hidden for a user if they don't have permission.

 > What happens if someone provides icons for some, but not all of the root items/groups?

I think it should be a requirement for someone to provide icons on the root items/groups as they are necessary for when the nav collapses. If there is no icon then there would be nothing to click on when it's collapsed. It would take a redesign of the collapsed nav in order for this to not become necessary.

PR #56, opened 2019-07-10
Go-Copy Button

StevenUlmer:
This allows a user to copy some text by clicking on a clipboard icon. Whatever text you pass into the component will get copied to the users clipboard when they click on the icon.


projects/go-lib/src/lib/components/go-copy/go-copy.component.scss
```diff
@@ -0,0 +1,4 @@
+.hideTextArea {
+  position: absolute;
+  left: 99999px;
```
StevenUlmer: I really hate this, but I'm not sure of a better way to do this as doing display:none or setting the height and width to 0 don't work.

projects/go-lib/src/lib/components/go-copy/go-copy.component.html
```diff
@@ -0,0 +1,4 @@
+<div (click)="copyStringToClipboard($event.currentTarget)">
```
AlexOverbeck: We need to remove the id from the textarea here so we avoid having duplicate ids on the page. We could also clean up our BEM selectors if we structure things like the following:
```html
<div class="go-copy">
  <go-icon icon="file_copy"></go-icon>
  <textarea class="go-copy__textarea">{{text}}</textarea>
</div>
```

projects/go-lib/src/lib/components/go-copy/go-copy.component.scss
```diff
@@ -0,0 +1,4 @@
+.hideTextArea {
```
AlexOverbeck: If we update the markup like in the previous comment this would become: `.go-copy__textarea`

projects/go-lib/src/lib/components/go-copy/go-copy.component.scss
AlexOverbeck: Yeah, that makes sense if we want it to work in various browsers.

projects/go-lib/src/lib/components/go-copy/go-copy.component.ts
```diff
@@ -0,0 +1,18 @@
...
+  selector: 'go-copy',
+  templateUrl: './go-copy.component.html',
+  styleUrls: ['./go-copy.component.scss']
+})
+export class GoCopyComponent {
+  @Input() text: string;
+
+  constructor() { }
+
+  copyStringToClipboard (element: HTMLInputElement) {
```
AlexOverbeck: There isn't a return type on this function. If we want to make the most of typescript we should add one:
```suggestion
  copyStringToClipboard(element: HTMLInputElement): void {
```

PR #57, opened 2019-07-10
Adds all files necessary for GoIconButtonComponent

jaredami:
Creates a component for buttons with icons that do not have text.

Implementation example:
```
<go-icon-button buttonIcon="home" (handleClick)="openThing()"></go-icon-button>
```

@grahamhency - review everything

### DIFF
--- a/projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
+++ b/projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
@@ -6,10 +6,10 @@
 
   border: 0;
   border-radius: $global-radius--round;
-  color: $base-dark;
+  color: $theme-light-color;
   cursor: pointer;
   display: inline-flex;
-  outline-color: $theme-light-border;
+  box-shadow: $theme-light-border;
   padding: .5rem;
 
   &:hover {
--- a/projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
+++ b/projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
@@ -4,12 +4,13 @@
 .go-icon-button {
   @include transition(background-color);
 
+  background: $theme-light-bg;
   border: 0;
   border-radius: $global-radius--round;
+  box-shadow: $global-box-shadow;
   color: $theme-light-color;
   cursor: pointer;
   display: inline-flex;
-  box-shadow: $theme-light-border;
   padding: .5rem;
 
   &:hover {
--- a/projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
+++ b/projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
@@ -1,0 +1,18 @@
+@import '~@tangoe/gosheets/base/variables';
+@import '~@tangoe/gosheets/base/mixins';
+
+.go-icon-button {
+  @include transition(background-color);
+
+  border: 0;
+  border-radius: $global-radius--round;
+  color: $base-dark;
+  cursor: pointer;
+  display: inline-flex;
+  outline-color: $theme-light-border;
+  padding: .5rem;
+
+  &:hover {
+    background-color: $theme-light-bg-hover;
+  }
+}
--- a/projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.ts
+++ b/projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.ts
@@ -8,7 +8,7 @@ export class GoIconButtonComponent {
   @Input() buttonDisabled: boolean;
   @Input() buttonIcon: string;
 
-  @Output() handleClick = new EventEmitter<boolean>();
+  @Output() handleClick = new EventEmitter();
 
   public clicked(): void {
     this.handleClick.emit();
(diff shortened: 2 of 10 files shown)


projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
```diff
@@ -0,0 +1,18 @@
+@import '~@tangoe/gosheets/base/variables';
+@import '~@tangoe/gosheets/base/mixins';
+
+.go-icon-button {
+  @include transition(background-color);
+
+  border: 0;
+  border-radius: $global-radius--round;
+  color: $base-dark;
```
grahamhency: Although they're the same color, we should probably use `$theme-light-color` for this.

projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
```diff
@@ -0,0 +1,18 @@
...
+  cursor: pointer;
+  display: inline-flex;
+  outline-color: $theme-light-border;
```
grahamhency: Not sure about this, would like to see what this looks like. We have a lot more control with a `box-shadow` so I would probably opt for that. Another question along these same lines is the interactivity of the button. Hover, focus, active states.

projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
```diff
@@ -0,0 +1,18 @@
...
+  padding: .5rem;
+
+  &:hover {
+    background-color: $theme-light-bg-hover;
+  }
+}
```
grahamhency: Just need a new line at the end of this file.

projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.ts
```diff
@@ -0,0 +1,16 @@
...
+
+@Component({
+  selector: 'go-icon-button',
+  templateUrl: './go-icon-button.component.html'
+})
+export class GoIconButtonComponent {
+  @Input() buttonDisabled: boolean;
+  @Input() buttonIcon: string;
+
+  @Output() handleClick = new EventEmitter<boolean>();
```
grahamhency: It looks like this `Output` should emit a type of `boolean`, however when `emit` is called down below on line `14` we're not passing anything in the params so it'll always be false. I don't think this _really_ matters, but it might be worthwhile to change this to emit a type of `void`. Thoughts?

projects/go-lib/src/lib/go-shared.module.ts
```diff
@@ -8,13 +8,15 @@ import { GoModalModule } from './components/go-modal/go-modal.module';
...
 import { GoToasterModule } from './components/go-toaster/go-toaster.module';
+import { GoIconButtonModule } from './components/go-icon-button/go-icon-button.module';
 
 @NgModule({
   imports: [
     GoAccordionModule,
     GoButtonModule,
     GoCardModule,
     GoIconModule,
+    GoIconButtonModule,
```
grahamhency: For the sake of consistency, could we alphabetize this file? This spot, the import above, and the exports below.

projects/go-tester/src/app/app.module.ts
```diff
@@ -40,7 +43,8 @@ import { AppService } from './app.service';
   ],
   entryComponents: [
     GoButtonComponent,
-    GoIconComponent
+    GoIconComponent,
+    GoIconButtonComponent
```
grahamhency: This won't affect the implementation of the button at all, but we actually don't need to add the component to the `entryComponents`. We only need to do this if we're injecting the `GoIconButtonComponent` into our modal or off-canvas. In this case, it doesn't seem like we are so we can probably remove the component from this array and actually remove the import for the component. **not the module though**

AlexOverbeck: I think this was for testing purposes. It might be helpful for all of us to get together and chat about this. I don't fully understand this within the context of this project yet.

projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
AlexOverbeck: Yeah, we should switch to using a `box-shadow` for this.

projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.ts
jaredami: Agreed!

projects/go-tester/src/app/app.module.ts
jaredami: I removed the component from the `entryComponents` after our offline discussion.

projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
```diff
@@ -0,0 +1,18 @@
...
+
+.go-icon-button {
+  @include transition(background-color);
+
+  border: 0;
+  border-radius: $global-radius--round;
+  color: $theme-light-color;
+  cursor: pointer;
+  display: inline-flex;
+  box-shadow: $theme-light-border;
```
grahamhency: I don't think this would generate a correct property value. I think you want `$global-box-shadow` here!

projects/go-lib/src/lib/components/go-icon-button/go-icon-button.component.scss
```diff
@@ -0,0 +1,18 @@
+@import '~@tangoe/gosheets/base/variables';
+@import '~@tangoe/gosheets/base/mixins';
+
+.go-icon-button {
```
grahamhency: We should probably throw a `background` attribute on here, just to enforce where this is used.
`background: $theme-light-bg;`

PR #64, opened 2019-07-11
[Feature] Go Badge

StevenUlmer:
This adds a badge component to be used with buttons and icons. If we need to use it outside of those components we should do some more testing around this as those are the only two things that I tested this with.

### DIFF
--- a/projects/go-lib/src/lib/components/go-badge/go-badge.component.html
+++ b/projects/go-lib/src/lib/components/go-badge/go-badge.component.html
@@ -1,0 +1,6 @@
+<span class="go-badge-outer-container">
+  <ng-content></ng-content>
+  <span class="go-badge-container" *ngIf="badgeData">
+    <span class="go-badge">{{badgeData}}</span>
+  </span>
+</span>
--- a/projects/go-lib/src/lib/components/go-badge/go-badge.component.html
+++ b/projects/go-lib/src/lib/components/go-badge/go-badge.component.html
@@ -1,6 +1,6 @@
 <span class="go-badge-outer-container">
   <ng-content></ng-content>
-  <span class="go-badge-container" *ngIf="badgeData">
-    <span class="go-badge" [ngClass]="getColor()">{{badgeData}}</span>
+  <span class="go-badge-container">
+    <span class="go-badge" [ngClass]="badgeClasses()">{{ displayData ? badgeData : ''}}</span>
   </span>
 </span>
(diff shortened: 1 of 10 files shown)


projects/go-lib/src/lib/components/go-badge/go-badge.component.html
```diff
@@ -0,0 +1,6 @@
+<span class="go-badge-outer-container">
+  <ng-content></ng-content>
+  <span class="go-badge-container">
+    <span class="go-badge" [ngClass]="badgeClasses()">{{ displayData ? badgeData : ''}}</span>
```
grahamhency: Should we consider not using `ngClass` here because we can probably assume that this is not going to change outside the scope of this component. I think we could make use of a local object that holds this value and then make use of `onChanges` to reset these values. That way we're only reevaluating these classes when changes to the inputs happen. Thoughts?

projects/go-lib/src/lib/components/go-badge/go-badge.component.scss
```diff
@@ -0,0 +1,47 @@
+@import '~@tangoe/gosheets/base/_variables';
+
+.go-badge {
+  border: 1px;
+  border-color: $theme-light-bg;
+  border-radius: 25px;
+  border-style: solid;
```
grahamhency: Should we just consolidate these into the short hand `border`?

projects/go-lib/src/lib/components/go-badge/go-badge.component.scss
```diff
@@ -0,0 +1,47 @@
...
+  font-size: 12px;
```
grahamhency: Can we make use of rems here?

projects/go-lib/src/lib/components/go-badge/go-badge.component.ts
```diff
@@ -0,0 +1,23 @@
...
+
+@Component({
+  selector: 'go-badge',
+  templateUrl: './go-badge.component.html',
+  styleUrls: ['./go-badge.component.scss']
+})
+export class GoBadgeComponent {
+  @Input() badgeData: string;
+  @Input() badgeColor: string = 'neutral';
+  @Input() displayData: boolean = true;
```
grahamhency: Should we set the `displayData` to false by default?

StevenUlmer: I think this should be true by default. The reason I say this is because most times you are adding a badge you would expect whatever data you are adding to that badge to show up. There are very few instances that I can think of where you would just want the badge without the data showing up and all of them are pretty conditional.

PR #65, opened 2019-07-11
Fix spec tests

AlexOverbeck:
### Purpose
1. Fixes all of our failing tests
2. Configures TravisCI to run those tests
3. Removes all the karma tools that we are using yet
4. Updates Karma configuration to reflect those changes


.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: node_js
+node_js:
+  - 8
```
StevenUlmer: Should we specify a more precise version of node here? We're using `"node": "^8.11.4"` in the package.json, should we use that same version here?

AlexOverbeck: Yeah, probably not a bad idea.

PR #70, opened 2019-07-19
Adds tests for go-icon component

AlexOverbeck:
This adds basic unit tests around the `<go-icon>` component and it's corresponding template.


projects/go-lib/src/lib/components/go-icon/go-icon.component.spec.ts
```diff
@@ -22,4 +22,56 @@ describe('GoIconComponent', () => {
...
+
+      component.iconClass = 'odd multiple class use case';
+
+      expect(component.classObject()).toEqual({
+        'odd multiple class use case': true
+      });
+    });
+  });
+
+  describe('the template', () => {
```
AlexOverbeck: Testing templates probably isn't something we need to do often, but it seemed so easy in this case that I figured we might as well.

PR #71, opened 2019-07-19
[Feature] Search Component

grahamhency:
## Source
https://mobi.teamwork.com/#tasks/13648494

## Description
This component is to be used in the header to provide a consistent, global, search.

## Screenshots
Screen Shot 2019-07-19 at 3 31 04 PM
Screen Shot 2019-07-19 at 3 31 39 PM

## Implementation

This should be rendered in the root component insides the `go-header`
```
<go-search>
  <!-- the component inside of the search is to be built by the implementer -->
  <search-dummy-results></search-dummy-results>
</go-search>
```

Component file for `<search-dummy-results>`:
```
this.searchService.searchTerm.subscribe(searchTerm => {

  // this section is dependent upon what the data looks like
  this.appService
        .getMockSearch(searchTerm)
        .subscribe(results => {
           if (results.length === 0) {
             this.results = null;
             this.searchService.notFoundResponse();
           } else {
             this.results = results;
             this.searchService.successResponse();
           }
        });
  });
}
```

### DIFF
--- a/projects/go-lib/src/lib/components/go-search/go-search.component.spec.ts
+++ b/projects/go-lib/src/lib/components/go-search/go-search.component.spec.ts
@@ -1,0 +1,36 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { ReactiveFormsModule } from '@angular/forms';
+import { CommonModule } from '@angular/common';
+
+import { GoSearchComponent } from './go-search.component';
+
+import { GoIconModule } from '../go-icon/go-icon.module';
+import { GoLoaderModule } from '../go-loader/go-loader.module';
+
+describe('GoSearchComponent', () => {
+  let component: GoSearchComponent;
+  let fixture: ComponentFixture<GoSearchComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ GoSearchComponent ],
+      imports: [
+        CommonModule,
+        GoIconModule,
+        GoLoaderModule,
+        ReactiveFormsModule
+      ]
+    })
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(GoSearchComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
--- a/projects/go-lib/src/lib/components/go-search/go-search.component.ts
+++ b/projects/go-lib/src/lib/components/go-search/go-search.component.ts
@@ -1,0 +1,82 @@
+import { Component, ElementRef, HostListener, OnInit } from '@angular/core';
+import { AnimationEvent } from '@angular/animations';
+import { FormBuilder, FormGroup } from '@angular/forms';
+import { debounceTime, distinctUntilChanged } from 'rxjs/operators';
+
+import { searchLoaderAnim, searchResultsAnim } from '../../animations/search.animation';
+import { GoSearchService } from './go-search.service';
+
+@Component({
+  selector: 'go-search',
+  templateUrl: './go-search.component.html',
+  styleUrls: ['./go-search.component.scss'],
+  animations: [searchLoaderAnim, searchResultsAnim]
+})
+export class GoSearchComponent implements OnInit {
+
+  goSearchForm: FormGroup;
+  searchActive: boolean = false;
+  resultsOverflow: string = 'hidden';
+
+  @HostListener('document:click') onDocumentClick(event) {
+    this.closeSearchEvent(event);
+  }
+
+  constructor(
+    public goSearchService: GoSearchService,
+    private elementRef: ElementRef,
+    private fb: FormBuilder
+  ) {
+    this.goSearchForm = this.fb.group({
+      term: ''
+    });
+  }
+
+  ngOnInit(): void {
+    this.goSearchForm.valueChanges.pipe(
+      debounceTime(500),
+      distinctUntilChanged()
+    ).subscribe(changes => {
+      if (changes['term'].length >= this.goSearchService.termLength) {
+        this.goSearchService.showNoResultsMessage = false;
+        this.goSearchService.showLoader = true;
+        this.goSearchService.updateSearchTerm(changes['term']);
+      } else {
+        this.goSearchService.showNoResultsMessage = false;
+        this.goSearchService.hasResults = false;
+        this.goSearchService.showLoader = false;
+      }
+    });
+  }
+
+  resultsStarted(event: AnimationEvent): void {
+    this.resultsOverflow = 'hidden';
+  }
+
+  resultsEnded(event: AnimationEvent): void {
+    this.resultsOverflow = event.toState === null ? 'auto' : 'hidden';
+  }
+
+  toggleActive(): void {
+    this.searchActive = true;
+  }
+
+  leaveInput(event: any): void {
+    if (!this.elementRef.nativeElement.contains(event.relatedTarget)) {
+      this.closeSearch();
+    }
+  }
+
+  closeSearchEvent(event: any): void {
+    if (event && !this.elementRef.nativeElement.contains(event.target)) {
+      this.closeSearch();
+    }
+  }
+
+  closeSearch(): void {
+    this.searchActive = false;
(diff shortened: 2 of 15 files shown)


projects/go-lib/src/lib/components/go-search/go-search.component.ts
```diff
@@ -0,0 +1,82 @@
...
+  styleUrls: ['./go-search.component.scss'],
+  animations: [searchLoaderAnim, searchResultsAnim]
+})
+export class GoSearchComponent implements OnInit {
+
+  goSearchForm: FormGroup;
+  searchActive: boolean = false;
+  resultsOverflow: string = 'hidden';
+
+  @HostListener('document:click') onDocumentClick(event) {
```
AlexOverbeck: HostListener doesn't appear to pass any arguments by default. You can configure and array of arguments to be passed to the `onDocumentClick` function as the second argument.

We can use this to pass the `$event.target` rather than the whole `$event` object, which is a [bad practice in Angular](https://angular.io/guide/user-input#passing-event-is-a-dubious-practice).

Consider this:
```javascript
@HostListener('document:click', ['$event.target'])
private onDocumentClick(target: HTMLElement): void {
  this.closeSearchEvent(target);
}
```

projects/go-lib/src/lib/components/go-search/go-search.component.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  closeSearchEvent(event: any): void {
+    if (event && !this.elementRef.nativeElement.contains(event.target)) {
+      this.closeSearch();
+    }
+  }
+
+  closeSearch(): void {
```
AlexOverbeck: This looks like a `private` function.

projects/go-lib/src/lib/components/go-search/go-search.component.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  resultsEnded(event: AnimationEvent): void {
+    this.resultsOverflow = event.toState === null ? 'auto' : 'hidden';
+  }
+
+  toggleActive(): void {
+    this.searchActive = true;
+  }
+
+  leaveInput(event: any): void {
```
AlexOverbeck: This `event` will always be a `FocusEvent`.
```suggestion
  leaveInput(event: FocusEvent): void {
```

projects/go-lib/src/lib/components/go-search/go-search.component.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (!this.elementRef.nativeElement.contains(event.relatedTarget)) {
+      this.closeSearch();
+    }
+  }
+
+  closeSearchEvent(event: any): void {
```
AlexOverbeck: If we make the above suggestion to the `@HostListener` this could become:
```typescript
private closeSearchEvent(eventTarget: HTMLElement): void {
  if (!this.elementRef.nativeElement.contains(eventTarget)) {
    this.closeSearch();
  }
}
```

projects/go-lib/src/lib/components/go-search/go-search.component.ts
```diff
@@ -0,0 +1,82 @@
...
+  resultsOverflow: string = 'hidden';
+
+  @HostListener('document:click') onDocumentClick(event) {
+    this.closeSearchEvent(event);
+  }
+
+  constructor(
+    public goSearchService: GoSearchService,
+    private elementRef: ElementRef,
+    private fb: FormBuilder
```
AlexOverbeck: Do we think calling this `fb` would be something that is easy to guess if this component got a little more lengthy and the variable was referenced further away from where it is declared?

I personally looked at line 30 and was like WTF is `fb`.

projects/go-lib/src/lib/components/go-search/go-search.component.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(GoSearchComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
```
AlexOverbeck: Tests will come soon.

projects/go-tester/src/app/app.component.html
```diff
@@ -6,6 +6,14 @@ <h1 class="go-header-2">
     </h1>
   </div>
 
+  <div class="go-column go-column--100">
+    <div style="height: 4rem; background: white; display: flex; align-items: center; padding: 0 1rem;">
```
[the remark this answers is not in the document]
grahamhency: @AlexOverbeck I was just doing this to imitate the header, but I can remove them.

projects/go-lib/src/lib/components/go-search/go-search.service.ts
```diff
@@ -0,0 +1,67 @@
...
+   */
+  noResultsMessage: string = 'No Results Found';
+
+  /**
+   * Whether or not to show the noResultsMessage
+   */
+  showNoResultsMessage: boolean = false;
+
+  /**
+   * Whether or not to show the loader in the search bar
```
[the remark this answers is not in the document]
grahamhency: @AlexOverbeck I think in this case, it's alright. These will show up in vscode when someone goes to implement the service as the descriptive documentation. These will not get included in the production build of the app, thus I believe it's acceptable to be more verbose about what each of these properties do than not.
Screen Shot 2019-07-23 at 8 44 03 AM

projects/go-lib/src/lib/components/go-search/go-search.component.ts
grahamhency: @AlexOverbeck interesting, I understand what they're getting at here. This is actually broke, lol. I implemented this approach and fixed the issue. Good catch.

projects/go-lib/src/lib/components/go-search/go-search.component.ts
grahamhency: Yep, good idea.

projects/go-lib/src/lib/components/go-search/go-search.component.ts
grahamhency: Yes! Changed it, thanks.

projects/go-lib/src/lib/components/go-search/go-search.component.ts
grahamhency: That's fair, I think it's common in Angular to name it that... but I've changed it as it doesn't hurt to be slightly more direct.

projects/go-lib/src/lib/components/go-search/go-search.component.ts
```diff
@@ -61,19 +62,19 @@ export class GoSearchComponent implements OnInit {
     this.searchActive = true;
   }
 
-  leaveInput(event: any): void {
-    if (!this.elementRef.nativeElement.contains(event.relatedTarget)) {
+  leaveInput(event: FocusEvent): void {
+    if (event.relatedTarget && !this.elementRef.nativeElement.contains(event.relatedTarget)) {
```
AlexOverbeck: What is the case where a FocusEvent doesn't have a related target? I guess if you just click off of something?

projects/go-tester/src/app/app.component.html
```diff
@@ -18,7 +18,7 @@ <h1 class="go-header-2">
 
         <div class="go-column go-column--100">
           <h4>Search</h4>
-          <div style="height: 4rem; background: white; display: flex; align-items: center; padding: 0 1rem;">
+          <div>
```
AlexOverbeck: But, where did the styles go?

grahamhency: @AlexOverbeck we don't need them, this is just the tester.

projects/go-lib/src/lib/components/go-search/go-search.component.ts
grahamhency: @AlexOverbeck exactly! This happens because the FocusEvent would be fired because a click off of the input, but not into another input. This causes an issue because `nativeElement.contains(null)` is true, apparently. Meaning that it would close the search if you clicked on the search results, which is not what we want. The other click event handler in this component handles closing it from a clicking off. This method should only handle the focus leave event.

PR #73, opened 2019-07-22
Test button component

AlexOverbeck:
#### Update go-tester karma config
This removes kjhtml from the karma config because that node module
has been removed.

#### Adds tests for <go-button> component
This adds basic unit tests around buttons. It additionally tests
some of the basic side effects that the data has on the go-button
template.

### DIFF
--- a/projects/go-lib/src/lib/components/go-button/go-button.component.spec.ts
+++ b/projects/go-lib/src/lib/components/go-button/go-button.component.spec.ts
@@ -32,7 +32,7 @@ describe('GoButtonComponent', () => {
       expect(component.isProcessing).toBe(component.useLoader);
     });
 
-    it('sets isProcessing to true if useLoader is true', () => {
+    it('sets isProcessing to useLoader', () => {
       component.useLoader =  true;
 
       expect(component.isProcessing).not.toBe(component.useLoader);
--- a/projects/go-lib/src/lib/components/go-button/go-button.component.spec.ts
+++ b/projects/go-lib/src/lib/components/go-button/go-button.component.spec.ts
@@ -29,15 +29,15 @@ describe('GoButtonComponent', () => {
     it('sets isProcessing to false if useLoader is false', () => {
       expect(component.isProcessing).toBe(false);
       component.clicked();
-      expect(component.isProcessing).toBe(false);
+      expect(component.isProcessing).toBe(component.useLoader);
     });
 
     it('sets isProcessing to true if useLoader is true', () => {
       component.useLoader =  true;
 
-      expect(component.isProcessing).toBe(false);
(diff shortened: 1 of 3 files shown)


projects/go-lib/src/lib/components/go-button/go-button.component.spec.ts
```diff
@@ -24,4 +24,145 @@ describe('GoButtonComponent', () => {
...
   });
+
+  describe('clicked()', () => {
+    it('sets isProcessing to false if useLoader is false', () => {
+      expect(component.isProcessing).toBe(false);
+      component.clicked();
+      expect(component.isProcessing).toBe(false);
+    });
+
+    it('sets isProcessing to true if useLoader is true', () => {
```
grahamhency: We could probably simplify this to just test that `isProcessing === useLoader` right?

projects/go-lib/src/lib/components/go-button/go-button.component.spec.ts
```diff
@@ -24,4 +24,145 @@ describe('GoButtonComponent', () => {
...
+    });
+
+    it('returns an object that set go-button--positive to true based on buttonVariant', () => {
+      expect(component.classObject()['go-button--positive']).toBe(false);
+
+      component.buttonVariant = 'positive';
+      expect(component.classObject()['go-button--positive']).toBe(true);
+    });
+
+    it('returns an object that set other modifiers to true', () => {
```
grahamhency: Should these be separated out into individual tests?

AlexOverbeck: Meh, I didn't really want to have a bunch of tests to test individual keys in the object. We could separate them in the future if we would like.

projects/go-lib/src/lib/components/go-button/go-button.component.spec.ts
grahamhency: I see you updated the assertions in this test to reflect `isProcessing === useLoader`, but should we update the description to reflect this updated test?

projects/go-lib/src/lib/components/go-button/go-button.component.spec.ts
```diff
@@ -24,4 +24,145 @@ describe('GoButtonComponent', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  describe('clicked()', () => {
+    it('sets isProcessing to false if useLoader is false', () => {
```
grahamhency: Do we need this test now that we've got the one below it?

AlexOverbeck: Yep, I wanted to test things in both cases.

PR #75, opened 2019-07-22
Cleans up and tests the go-loader component

AlexOverbeck:


projects/go-lib/src/lib/components/go-loader/go-loader.component.ts
```diff
@@ -11,30 +11,25 @@ export class GoLoaderComponent implements OnInit {
...
   ngOnInit() {
-  }
-
-  completeAnimation() : void {
-    this.loaderDone = true;
-  }
-
-  public pathClasses() : object {
-    return {
+    this.pathClasses = {
```
grahamhency: @AlexOverbeck I don't know if this will trigger these to update if the `@Input` is changed. Before, `[ngClass]` was calling a function which Angular will fire again if changes are made. In this approach, `pathClasses` are set `onInit` and that's it. Then the `[ngClass]` only knows to look at `pathClasses` which is a local variable and shouldn't be re-evaluated with a change event, unless specifically told to. Thoughts?

AlexOverbeck: Yeah. that's true. At the time I was thinking that there wouldn't be a use case where we need to change the inputs after the component initialized.

PR #76, opened 2019-07-22
Updates npm dependency to fix vulnerabilities

AlexOverbeck:


AlexOverbeck:
@StevenUlmer I don't think so. This doesn't have much to do with the actual code that is being released.

PR #77, opened 2019-07-23
Updates Button Loading Spinner

### LINKED ISSUE #68: <go-button> does not submit form when useLoader is true
### Current Behavior 🐛
When setting the `[useLoader]` input on the GoButtonComponent, the click and submit events are not be propagated to the form, and any submit handler attached to the form will not be fired. The only way is to bind to the `(handleClick)` on the go-button, which is fine, but weird to be constrained to that.

This works 🤓
```html
<form #f="ngForm">  <!-- form submission only being handled by the button below 🤔 -->
  <!-- here be inputs to gather super important info-->
  <go-button
     buttonType="submit"
     (handleClick)="onSubmit(f)"
     [useLoader]="true"
      #submitButton
  >
    Submit
  </go-button>
</form>
```
This does not work 😞
```html
<form #f="ngForm" (ngSubmit)="onSubmit(f)"> <!-- submit is not being called 😿 -->
  <!-- here be inputs to gather super important info -->
  <go-button
     buttonType="submit"
     [useLoader]="true"
      #submitButton
  >
    Submit
  </go-button>
</form>
```

### Expected Behavior 🌻
Setting the loading status should not interfere with the button's good ol behavior of submitting the form.

### Possible Solution 💡
No idea cuz not sure exactly where the bug is. I do see that the button gets disabled when `useLoader` is true, but not sure why that should interfere with the form submission.

StevenUlmer:
Fixes #68
The way our button currently works we are setting the loader on the button itself. The problem with this is it disables the button before the js triggers that a form submit has been clicked. This causes the Form to never be submitted. The only ways I found around this are setting a timeout before we disable the button or to completely remove the disabling of the button with the loading spinner from the go button itself and put that responsibility on the parent component instead. The parent would update whether the button is processing using a binding and when the button is processing it would disable the button and add the loading spinner. This does change how we were using the loading spinner before tho and would be a breaking change to any buttons that currently use the loading spinner.


projects/go-lib/src/lib/components/go-button/go-button.component.ts
```diff
@@ -9,24 +9,17 @@ export class GoButtonComponent {
   @Input() buttonIcon: string;
   @Input() buttonType: string = 'button';
   @Input() buttonVariant: string;
-  @Input() useLoader: boolean;
```
StevenUlmer: we could make this a little less breaking by keeping the binding here and just updating the html to use `useLoader` instead of isProcessing

projects/go-lib/src/lib/components/go-button/go-button.component.ts
```diff
@@ -9,21 +9,18 @@ export class GoButtonComponent {
   @Input() buttonIcon: string;
   @Input() buttonType: string = 'button';
   @Input() buttonVariant: string;
-  @Input() useLoader: boolean = false;
+  @Input() isProcessing: boolean = false;
   @Input() useDarkTheme: boolean = false;
+  @Input() useLoader: boolean = false;
```
StevenUlmer: Leaving this here, so that it won't give compiler errors when introducing this change to an app that already uses it. In a later version we can completely deprecate it. This way it gives people time to get away from it instead of just breaking all of the templates that use it

PR #78, opened 2019-07-23
Go Toast Tests

StevenUlmer:
This adds basic unit tests around the <go-toast> component and it's corresponding template.

### DIFF
--- a/projects/go-lib/src/lib/components/go-toast/go-toast.component.spec.ts
+++ b/projects/go-lib/src/lib/components/go-toast/go-toast.component.spec.ts
@@ -18,10 +18,77 @@ describe('GoToastComponent', () => {
   beforeEach(() => {
     fixture = TestBed.createComponent(GoToastComponent);
     component = fixture.componentInstance;
-    fixture.detectChanges();
   });
 
   it('should create', () => {
     expect(component).toBeTruthy();
+  });
+
+  describe('ngOnInit', () => {
+    it('sets a default statusClass if no type is passed in', () => {
+      fixture.detectChanges();
+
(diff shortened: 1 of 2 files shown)


projects/go-lib/src/lib/components/go-toast/go-toast.component.spec.ts
```diff
@@ -18,10 +18,77 @@ describe('GoToastComponent', () => {
...
+      buttonElement.dispatchEvent(new Event('click'));
+      fixture.detectChanges();
+
+      expect(component.dismiss).toHaveBeenCalled();
+      expect(component.handleDismiss.emit).toHaveBeenCalled();
+    });
+  });
+
+  describe('the template', () => {
+    it('renders dismiss button if dismissable is true', () => {
```
grahamhency: Does this test pass?
`dismissable` is false be default, so to test this case correctly I think the assertion needs to be:
`expect(goToastDismissButton).not.toBeNull();`

projects/go-lib/src/lib/components/go-toast/go-toast.component.spec.ts
```diff
@@ -18,10 +18,77 @@ describe('GoToastComponent', () => {
...
+      fixture.detectChanges();
+
+      const goToastTemplate: HTMLElement = fixture.nativeElement;
+      const goToastDismissButton: HTMLElement = goToastTemplate.querySelector('.go-toast-dismiss__button');
+
+      expect(goToastDismissButton).toBeNull();
+    });
+
+    it('hides the dismiss button by default', () => {
```
grahamhency: The logic for this one makes sense, but it's the opposite of the `it`. Might need to address this description

Issue #59, opened 2019-07-11
Create a `<go-global-header>` component


AlexOverbeck:
This represents one of the more challenging pieces of the app shell project. We need a way to allow some flexibility between products. For example, one product might have a "new activity" or "user profile" button that I assume not all of the other companies have. The way I see it we could just allow for three areas of templating within this component. A left, middle, and right side template would allow each company to implement these areas as they would like while still giving us most of the control over the layout.

I would imagine something like this:
```html
<header class="go-global-header">
  <div class="go-global-header__logo">
    <ng-content select="[go-header-logo]"></ng-content>
  </div>
  <div class="go-global-header__search">
    <ng-content select="[go-header-search]"></ng-content>
  </div>
  <div class="go-global-header__actions">
    <ng-content select="[go-header-actions]"></ng-content>
  </div>
</header>
```
This would give us the flexibility to shift things around if we need to. It would let whoever is implementing these pieces deal with the complexities of implementing something like a global search component.

Issue #58, opened 2019-07-11
Create a `<go-layout>` component.


AlexOverbeck:
This component serves as the main app component. The `<go-layout>` component is a simple wrapper that includes all of the main components needed to get our UI up of the ground. We may or may not need this component, but it could be a useful way of getting things up and running quickly and consistently.

PR #79, opened 2019-07-23
[Feature] Go Layout

grahamhency:
## Description
This PR introduces two new components that rely on each other to form an application wrapper.

- Layout Component (fixes: #58)
- Header Component (fixes: #59)

## Screenshot
Screen Shot 2019-07-23 at 2 14 39 PM

### app.module.ts
```
import {
  GoHeaderModule,
  GoLayoutModule
} from '@tangoe/goponents';

@NgModule({
  imports: [
    GoHeaderModule
    GoLayoutModule
  ]
})
export class AppModule { }
```

### app.component.html
```
<go-layout>
  <go-header go-layout-header [logo]="logo">
    <go-search go-header-middle>
      <!-- this is where each product's search component will reside -->
      <app-search-test></app-search-test>
    </go-search>
    <ng-container go-header-right>
    <!-- this is where the actions on the right will reside -->
      <go-icon-button buttonIcon="picture_in_picture" (handleClick)="openModal()"></go-icon-button>
      <go-icon-button buttonIcon="input" (handleClick)="openOffCanvas()"></go-icon-button>
    </ng-container>
  </go-header>
  <go-side-nav go-layout-nav [menuItems]="menuItems">
  </go-side-nav>
</go-layout>
```


projects/go-lib/src/lib/components/go-header/go-header.component.html
```diff
@@ -0,0 +1,20 @@
+<header class="go-header">
+  <div class="go-header__left"
+       [ngClass]="{ 'go-header__left--collapsed': !sideNavService.navOpen }">
+    <go-icon class="go-header__menu"
+             icon="menu" 
+             (click)="toggleSideMenu()">
+    </go-icon>
+    <div class="go-header__logo-container"
+         [ngClass]="{ 'go-header__logo-container--collapsed': !sideNavService.navOpen }">
+      <img class="go-header__logo" [src]="logo"/>
```
StevenUlmer: Should we have an alt text here on this image? What happens if an invalid src or no src gets passed in?

projects/go-lib/src/lib/components/go-header/go-header.component.ts
```diff
@@ -0,0 +1,19 @@
+import { Component, Input } from '@angular/core';
+import { GoSideNavService } from '../go-side-nav/go-side-nav/go-side-nav.service';
+
+@Component({
+  selector: 'go-header',
+  templateUrl: './go-header.component.html',
+  styleUrls: ['./go-header.component.scss']
+})
+export class GoHeaderComponent {
+
```
StevenUlmer: extra white space here.

projects/go-lib/src/lib/components/go-layout/go-layout.component.ts
```diff
@@ -0,0 +1,48 @@
...
+  routeLoader: boolean = false;
+
+  constructor(private router: Router) { }
+
+  ngOnInit(): void {
+    this.router.events.subscribe(event => {
+      if (event instanceof NavigationStart) {
+        this.routeLoader = true;
+      }
+      if ((event instanceof NavigationEnd) || (event instanceof NavigationCancel) || (event instanceof NavigationError)) {
```
StevenUlmer: Should this be an else if? It shouldn't matter from a programming point of view as they don't overlap at all, but it does read a little easier for someone coming along later.

projects/go-tester/src/app/app.component.html
```diff
@@ -1,110 +1,13 @@
...
-        </div>
-      </div>
-    </div>
-  </div>
-</div>
-<go-toaster></go-toaster>
-<go-off-canvas></go-off-canvas>
+<go-layout>
+  <go-header go-layout-header [logo]="logo">
+    <go-search go-header-middle>
```
StevenUlmer: Are we going to run into problems here if a user doesn't have permission to view the search bar in the middle? Essentially if you put an ng-if on this go-header-middle will the header still look correct?

projects/go-tester/src/app/app.guard.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable } from '@angular/core';
+import { Router, CanActivate, ActivatedRouteSnapshot, RouterStateSnapshot } from '@angular/router';
+
+import { Observable } from 'rxjs';
+
+@Injectable()
+export class AppGuard implements CanActivate {
+
+	constructor(private router: Router) { }
```
StevenUlmer: can we look into formatting this file?

projects/go-lib/src/lib/components/go-header/go-header.component.html
grahamhency: Good point, I added a binding and output the `alt` here. :+1:

projects/go-lib/src/lib/components/go-layout/go-layout.component.ts
grahamhency: @StevenUlmer I suppose that's fair. I changed it to `else if`. I only did it that way because I felt like it was two separate bits of functionality. One was showing the loader, one was hiding, but it doesn't matter either way and `else if` probably is easier.

projects/go-tester/src/app/app.component.html
grahamhency: @StevenUlmer is that a thing? There are users that can't see the search bar? That being said, if the `go-layout` is rendered without a `go-header-middle` the header will just render with a blank spot in the middle like so:
Screen Shot 2019-07-24 at 8 57 00 AM

projects/go-tester/src/app/app.guard.ts
grahamhency: I suppose, it's totally for testing so... I fixed it though.

projects/go-tester/src/app/app.component.html
StevenUlmer: 👍 ya there's actually a large part of our users who can't see the search as they don't have permission to do anything with anyone besides themselves. This looks great tho without it 🎉

projects/go-lib/src/lib/components/go-header/go-header.component.html
```diff
@@ -0,0 +1,22 @@
+<header class="go-header">
+  <div class="go-header__left"
+       [ngClass]="{ 'go-header__left--collapsed': !sideNavService.navOpen }">
+    <go-icon class="go-header__menu"
```
AlexOverbeck: This should be wrapped in a `<button>` tag where the click action is on the button.

StevenUlmer: should we just use a go-icon-button here instead?

grahamhency: @StevenUlmer yes, good idea. I've updated this.

PR #89, opened 2019-07-24
Firefox Search Animation Errors

### LINKED ISSUE #81: Bug: Search loading animation in Firefox
## Problem
In Firefox, the search loading animation does not disappear when the results are loaded.

Screen Shot 2019-07-23 at 4 18 16 PM

StevenUlmer:
Fixes #81 

In firefox the loading spinner and no results content never go away after they appear. If they get triggered again, the content just gets duplicated. This is happening because of https://github.com/angular/angular/issues/20489. In order to fix it we can just add the base styles on the leave animation instead of relying on the browser to find them.


grahamhency:
@StevenUlmer I can confirm that this fixes the same exact issue in IE, 2 🐦 with 1 🗿

PR #91, opened 2019-07-25
Bug ie edge cases

grahamhency:
## Problem
Many IE bugs

## Solution
Squashed them all

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1440,7 +1440,6 @@
       "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.5.tgz",
       "integrity": "sha512-5hYdAkZlcG8tOLujVDTgCT+uPX0VnpAH28gWsLfzpXYm7wP6mp5Q/gYyR7YQ0cKVJcXJnl3j2kpBan13PtQf6w==",
       "dev": true,
-      "optional": true,
       "requires": {
         "delegates": "^1.0.0",
         "readable-stream": "^2.0.6"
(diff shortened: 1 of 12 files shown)


projects/go-tester/browserslist
```diff
@@ -8,4 +8,4 @@
 last 2 versions
 Firefox ESR
 not dead
-not IE 9-11
\ No newline at end of file
+IE 9-11
```
AlexOverbeck: Do we need to specify `9-11` here or can we just do `11`?

projects/go-tester/src/polyfills.ts
```diff
@@ -19,20 +19,20 @@
...
+import 'core-js/es6/parse-float';
+import 'core-js/es6/number';
+import 'core-js/es6/math';
+import 'core-js/es6/string';
+import 'core-js/es6/date';
+import 'core-js/es6/array';
+import 'core-js/es6/regexp';
+import 'core-js/es6/map';
+import 'core-js/es6/weak-map';
+import 'core-js/es6/set';
```
AlexOverbeck: This is kind of a bummer.

package-lock.json
```diff
@@ -1440,7 +1440,6 @@
       "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.5.tgz",
```
AlexOverbeck: Looks like you maybe used the wrong version of npm to install these modules. You shouldn't have this big of a diff.

projects/go-tester/browserslist
grahamhency: Not sure, I was just doing this for testing purposes and would be dependent upon the implementor to do. I bet we can just do `IE 11`

PR #93, opened 2019-07-25
Fixes issue with loader svg in Safari. Closes #82

### LINKED ISSUE #82: Bug: Go Loader does not work properly in Safari
## Problem
In Safari, the `go-loader` component renders solid black.

Screen Shot 2019-07-23 at 4 19 48 PM

AlexOverbeck:
~~There was an issue in the Safari browser where the loading svg~~
~~would look like three black circles. This was a result of there~~
~~being duplicate IDs within the svgs markup.~~

~~Generating unique IDs and applying the fill to the paths inline~~
~~based on those unique IDs fixes this problem.~~

The previous version of the loading svg had some issues in specific
browsers because it was reliant on gradient filters.

This redesigns the svg to no longer use gradients and uses angular 
to do some of the animations on the host element.


projects/go-lib/src/lib/components/go-loader/go-loader.component.scss
```diff
@@ -1,119 +1,67 @@
...
+  display: inline-block;
+  height: 75px;
+  position: relative;
+  width: 75px;
 
-.go-loader-gradient--start-negative {
-  stop-color: $ui-color-negative;
+  path {
+    fill: $ui-color-neutral;
+    transition: fill .5s;
```
grahamhency: Should we use the `transition` mixin here?

projects/go-lib/src/lib/components/go-loader/go-loader.component.scss
```diff
@@ -1,119 +1,67 @@
...
+.go-loader--small {
+  height: 50px;
+  width: 50px;
 }
 
-.go-loader--middle {
-	animation: rotatePathMiddle 1s linear infinite;
-	transform: translate(10%, 10%);
+.go-loader--medium {
+  height: 100px;
```
grahamhency: It looks like `go-loader` is set to dims of 75px x 75px, should we just set it to whatever the medium dims are? Thus, probably should change `go-loader` to 100px x 100px.

projects/go-lib/src/lib/components/go-loader/go-loader.component.ts
```diff
@@ -1,40 +1,33 @@
...
+export class GoLoaderComponent {
   @Input() loaderSize: string = 'medium';
   @Input() loaderType: string = 'neutral';
 
-  constructor() { }
-
-  ngOnInit() {
-  }
+  @HostBinding('@fadeTemplate')
+  public fadeTemplate: boolean = true;
```
grahamhency: I don't _think_ we really need to assign a variable to this `@HostBinding`.

projects/go-lib/src/lib/components/go-loader/go-loader.component.ts
```diff
@@ -1,40 +1,33 @@
...
+  animations: [
+    fadeTemplateAnimation
+  ]
 })
-export class GoLoaderComponent implements OnInit {
-
-  @Input() loaderDone: boolean = false;
+export class GoLoaderComponent {
   @Input() loaderSize: string = 'medium';
   @Input() loaderType: string = 'neutral';
```
grahamhency: We could probably throw some jsdoc comments on these variables to denote what options are available. Not necessary for this pr, but just thought about it.

projects/go-lib/src/lib/components/go-loader/go-loader.component.ts
```diff
@@ -1,40 +1,33 @@
...
+  @HostBinding('@fadeTemplate')
+  public fadeTemplate: boolean = true;
 
-  completeAnimation() : void {
-    this.loaderDone = true;
-  }
+  //#region Public Methods
 
-  public pathClasses() : object {
+  containerClasses(): object {
```
grahamhency: We probably don't need to call this `containerClasses` anymore because the `go-loader__container` class was removed. Now these are just `loaderClasses`. Not a necessary change, of course. I'll leave it up to you.

projects/go-lib/src/lib/components/go-loader/go-loader.component.ts
AlexOverbeck: Shrug, seems like it could be a nice override.

projects/go-lib/src/lib/components/go-loader/go-loader.component.scss
```diff
@@ -1,119 +1,53 @@
...
-  width: 100px;
-}
-
-.go-loader-container--large {
-  height: 150px;
-  width: 150px;
-}
+$loader-sizes: (small: 50px, medium: 100px, large: 150px);
+$loader-speeds: (outer: 1.25s, middle: 1s, inner: .75s);
+$loader-colors: (
```
grahamhency: These variables are a good idea, good thinkin!

PR #95, opened 2019-07-29
Updates linter rules to enforce type definitions everywhere

AlexOverbeck:
This updates out linter rules to enforce type definitions on all of the things.

---------------------
36rlfe


AlexOverbeck:
@StevenUlmer nope, this is not a part of that scope of work.

PR #128, opened 2019-08-08
Go Input Component

### LINKED ISSUE #119: Create a <go-input> component
We need to create a componentized version of a standard input to make it easier to control implementation and features for the input element.

2. Have a input with a type of text
3. Have a place to put a hint text to add additional information around the input
4. Handle its own errors with the addition of an error hint text.
5. Be able to be disabled
6. Support both a light and a dark theme
7. Be set up to work with Angular's [reactive forms](https://angular.io/guide/reactive-forms)

StevenUlmer:
Closes #119 

This adds an input component with a few bindings to allow for the use within reactive forms. It allows you to pass in both hints and errors to be displayed. It puts the responsibility of validation on the parent component allowing the implementor to have full control over the input.

### DIFF
--- a/projects/go-lib/src/lib/components/go-input/go-input.component.html
+++ b/projects/go-lib/src/lib/components/go-input/go-input.component.html
@@ -1,0 +1,22 @@
+<div [formGroup]="parentFormGroup">
+  <label [for]="controlName" class="go-form__label">
+    {{ label }}
+  </label>
+  <input [id]="controlName"
+         class="go-form__input"
+         [ngClass]="{'go-form__input--error': !controlIsValid()}"
+         [type]="inputType"
+         [placeholder]="placeholder"
+         [formControlName]="controlName">
+
+  <p *ngFor="let hint of hints" class="go-hint">
+    {{ hint }}
+  </p>
+
+  <ng-container *ngIf="!controlIsValid()">
+    <p class="go-hint go-hint--error" *ngFor="let error of errors">
+      <span class="go-hint__status">{{ errorStatus }}</span>
+      {{ error }}
+    </p>
+  </ng-container>
+</div>
--- a/projects/go-lib/src/lib/components/go-input/go-input.component.html
+++ b/projects/go-lib/src/lib/components/go-input/go-input.component.html
@@ -1,22 +1,25 @@
 <div [formGroup]="parentFormGroup">
-  <label [for]="controlName" class="go-form__label">
+  <label [for]="controlName"
+         class="go-form__label"
+         [ngClass]="{'go-form__label--dark': theme === 'dark'}">
     {{ label }}
   </label>
   <input [id]="controlName"
          class="go-form__input"
-         [ngClass]="{'go-form__input--error': !controlIsValid()}"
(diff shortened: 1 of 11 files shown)


projects/go-lib/src/lib/components/go-input/go-input.component.html
```diff
@@ -1,22 +1,27 @@
 <div [formGroup]="parentFormGroup">
-  <label [for]="controlName" class="go-form__label">
+  <label [for]="controlName"
+         class="go-form__label"
+         [ngClass]="{'go-form__label--dark': theme == 'dark'}">
```
grahamhency: Should we use `===` here?

projects/go-lib/src/lib/components/go-input/go-input.component.html
```diff
@@ -1,22 +1,27 @@
...
     {{ label }}
   </label>
   <input [id]="controlName"
          class="go-form__input"
-         [ngClass]="{'go-form__input--error': !controlIsValid()}"
+         [ngClass]="{'go-form__input--error': !controlIsValid(), 'go-form__input--dark': theme == 'dark'}"
```
grahamhency: Same here, `===`?

projects/go-lib/src/lib/components/go-input/go-input.component.html
```diff
@@ -1,22 +1,27 @@
...
          [type]="inputType"
          [placeholder]="placeholder"
          [formControlName]="controlName">
 
-  <p *ngFor="let hint of hints" class="go-hint">
+  <p *ngFor="let hint of hints" class="go-hint" [ngClass]="{'go-form__input--dark': theme == 'dark'}">
     {{ hint }}
   </p>
 
   <ng-container *ngIf="!controlIsValid()">
```
grahamhency: Should we apply this `ngIf` to the div below it to cut out an unnecessary html element?

projects/go-lib/src/lib/components/go-input/go-input.component.ts
```diff
@@ -7,14 +7,16 @@ import { FormGroup } from '@angular/forms';
...
   @Input() hints: string[];
-  @Input() controlName: string;
+  @Input() inputDisabled: boolean;
+  @Input() inputType: string = 'text';
+  @Input() label: string;
   @Input() parentFormGroup: FormGroup;
   @Input() placeholder: string = '';
+  @Input() theme: string = 'light';
 
   constructor() { }
```
grahamhency: Should we throw some errors in the `constructor` if any of the required items are not passed? Just a thought here, but we could do something like:
```
constructor() {
  if (!this.controlName) { throw new Error('GoInput: controlName is a required Input'); }
}
```

StevenUlmer: I could have done it wrong, but when I put this in the constructor it was throwing errors even when passing in the control name as it hadn't been defined yet. I went ahead and added it to the onInit tho

grahamhency: @StevenUlmer, my bad, I meant `onInit`. Thanks though, I think that'll help a lot.

PR #131, opened 2019-08-12
Go Textarea Component

### LINKED ISSUE #120: Create a <go-textarea> component
We need to create a componentized version of a textarea to make it easier to control implementation and features for the textarea element.

2. Have a textarea input
3. Have a place to put a hint text to add additional information around the input
4. Handle its own errors with the addition of an error hint text.
5. Be able to be disabled
6. Support both a light and a dark theme
7. Be set up to work with Angular's [reactive forms](https://angular.io/guide/reactive-forms)

StevenUlmer:
Closes #120

This adds a textarea component with a few bindings to allow for the use within reactive forms. It allows you to pass in both hints and errors to be displayed. It puts the responsibility of validation on the parent component allowing the implementor to have full control over the textarea.

### DIFF
--- a/projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
+++ b/projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
@@ -1,22 +1,26 @@
 <div [formGroup]="parentFormGroup">
-  <label [for]="controlName" class="go-form__label">
+  <label [for]="controlName"
+         class="go-form__label"
+         [ngClass]="{'go-form__label--dark': theme === 'dark'}">
     {{ label }}
   </label>
   <textarea [id]="controlName"
             class="go-form__input"
-            [ngClass]="{'go-form__input--error': !controlIsValid(), 'go-form__input--disabled': inputDisabled}"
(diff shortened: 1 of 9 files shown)


projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
```diff
@@ -0,0 +1,28 @@
+<div [formGroup]="parentFormGroup">
+  <label [for]="controlName"
+         class="go-form__label"
+         [ngClass]="{'go-form__label--dark': theme == 'dark'}">
```
grahamhency: Should we make this `===` ? Applies to the whole template.

projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
```diff
@@ -0,0 +1,28 @@
...
+            [placeholder]="placeholder"
+            [formControlName]="controlName"></textarea>
+
+  <p *ngFor="let hint of hints"
+     class="go-hint"
+     [ngClass]="{'go-form__input--dark': theme == 'dark'}">
+    {{ hint }}
+  </p>
+
+  <ng-container *ngIf="!controlIsValid()">
```
grahamhency: Probably could apply this `ngIf` to the `div` below it.

Issue #116, opened 2019-08-06
Table row selection


AlexOverbeck:
### What is the feature that you would like to propose?
We need to give our users a way to select specific rows of data from the **TableComponent**. This will allow them to subsequently perform an action on the selected data.

### Describe the solution you'd like
Lets present users with the ability to show a column of checkboxes in the first column. This feature should provide the following requirements.

1. Conditional row selection
    _During implementation we should be able to decide if a table has selectable rows._
2. Row selection
    _Users should be able to select a single row of data._
3. Multiple row section
    _User should be able to bulk select rows if they click the checkbox in the header._
4. Perform actions with selected data
    _Some form of event should be fired off so that developers can perform an action with the selected data._

PR #132, opened 2019-08-13
Feature table row selection

grahamhency:
## Description
This PR introduces the ability for table row selection in the GoTable.

Closes: #116 

### component.ts
```
@ViewChild('peopleTable') peopleTable: GoTableComponent;

        selectable: true,
        selectBy: 'id',
        tableData: results,
        totalCount: results.length
      });
      this.tableLoading = false;

outputSelectionCount(): void {
  console.log(this.peopleTable.getSelectionCount());
}

showCurrentSelection(): void {
  console.log(this.peopleTable.getSelectionState());
}

handleSelection(selectionEvent: RowSelectionEvent): void {
 console.log(selectionEvent);
}
```
### component.html
```
<go-table [tableConfig]="tableConfig"
          (rowSelectionEvent)="handleSelection($event)"
          [showTableActions]="true"
          #peopleTable
          *ngIf="tableConfig">
  <ng-container go-table-actions>
    <go-button (handleClick)="outputSelectionCount()">Selection Count</go-button>
    <go-button (handleClick)="showCurrentSelection()">Selection State</go-button>
  </ng-container>
  <go-table-column field="id" title="ID"></go-table-column>
  <go-table-column field="name.first" title="First Name"></go-table-column>
  <go-table-column field="name.last" title="Last Name"></go-table-column>
  <go-table-column field="email" title="Email"></go-table-column>
  <go-table-column field="gender" title="Gender"></go-table-column>
  <go-table-column field="ip_address" title="IP Address"></go-table-column>
</go-table>
```

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
@@ -1,0 +1,14 @@
+export interface GoTableRowSelectionEvent {
+  currentRow: {
+    data: object;
+    selected: boolean;
+  };
+  deselectedRows?: any[];
+  selectionMode: GoTableSelectionMode;
+  selectedRows?: any[];
+}
+
+export enum GoTableSelectionMode {
+  selection,
+  deselection
+}
--- a/projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
@@ -1,14 +1,41 @@
-export interface GoTableRowSelectionEvent {
+export interface RowSelectionEvent extends SelectionState {
+  /**
+   * The current row that was targeted for selection
+   */
   currentRow: {
+    /**
+     * The entire data in the row
+     */
     data: object;
+    /**
+     * Whether or not this row was selected (`true`) or deselected (`false`)
+     */
     selected: boolean;
   };
+}
+
+export interface SelectionState {
+  /**
+   * The current deselected rows; if `selectionMode` is `selection` then this will always be an empty array
+   */
   deselectedRows?: any[];
-  selectionMode: GoTableSelectionMode;
+
+  /**
+   * Defines the current mode of selection. `selection` will add items to `selectedRows`
+   * and `deselection` will add items to `deselectedRows`.
+   *
+   * If in `deselection` mode, the UI toggle all checkboxes to true. Items instead will be added to `deselectedRows`
+   * because they're being removed from the selection.
+   */
+  selectionMode: SelectionMode;
+
+  /**
+   * The current selected rows; if `selectionMode` is `deselection` then this will always be an empty array
+   */
   selectedRows?: any[];
 }
 
-export enum GoTableSelectionMode {
+export enum SelectionMode {
   selection,
   deselection
 }
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.ts
@@ -55,8 +55,7 @@ export class GoTableComponent implements OnInit, OnChanges {
 
   localTableConfig: GoTableConfig;
   selectAllChecked: boolean = false;
-  selectedRows: any[] = [];
-  deselectedRows: any[] = [];
+  targetedRows: any[] = [];
   showTable: boolean = false;
 
   ngOnInit(): void {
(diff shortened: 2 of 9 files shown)


projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
```diff
@@ -0,0 +1,41 @@
...
+   */
+  selectionMode: SelectionMode;
+
+  /**
+   * The current selected rows; if `selectionMode` is `deselection` then this will always be an empty array
+   */
+  selectedRows?: any[];
+}
+
+export enum SelectionMode {
```
AlexOverbeck: Is an enum the best way we can think of to store this selection mode? It doesn't seem very descriptive if I were to receive `1` or a `0` on the backend.

If we do decide this is the most simple way of doing this, do we want "selecting" to be a `0|false` or a `1|true`?

What are our other potential options? Maybe:
1. Storing this as a string?
    ```ruby
    {
      selectionMode: 'selecting' | 'deselecting'
    }
    ```
2. separating those out into two pieces of data?
    ```ruby
    {
      selecting: true | false,
      deselecting: true | false
    }
    ```
3. Maybe flipping the enum so the positives match up?
    ```javascript
    expect(selectionMode.deselection).toBe(0);
    expect(selectionMode.selection).toBe(1);
    ```

projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
```diff
@@ -0,0 +1,41 @@
+export interface RowSelectionEvent extends SelectionState {
+  /**
```
AlexOverbeck: All this documentation is super helpful.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -170,39 +216,124 @@ export class GoTableComponent implements OnInit, OnChanges {
...
     if (!this.isServerMode()) {
       this.loadingData = false;
     }
   }
+
+  private determineSelectionMode(): SelectionMode {
+    return this.selectAllChecked ? SelectionMode.deselection : SelectionMode.selection;
+  }
+
+  private isRowInTargeted(row: any): boolean {
```
AlexOverbeck: We always know what row will be right? It won't be `any` thing.

projects/go-tester/src/app/app.service.ts
```diff
@@ -5,18 +5,19 @@ import { GoTableConfig, GoTablePageConfig, GoTableSortConfig } from '../../../go
...
 
-  constructor (private http: HttpClient) { }
+  constructor(private http: HttpClient) { }
 
-  getMockData(params?: GoTableConfig) {
-    return this.http.get<any>("../assets/MOCK_DATA_1000.json").pipe(map(data => {
-      let formattedData = { totalCount: 0, results: [] };
+  getMockData(params?: GoTableConfig): Observable<any> {
+    return this.http.get<any>('../assets/MOCK_DATA_1000.json').pipe(map((data: any) => {
+      const formattedData: any = { totalCount: 0, results: [] };
```
AlexOverbeck: Lots of `any` in this file. Do we actually not know what type of things these are?

projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
grahamhency: @AlexOverbeck interesting, didn't really think about that. There are a couple approaches we can go with while still utilizing the `enum` as well to consider:

1. You can "stringify" the `enum` to the value of the enum and could send that to the backend, ex:
```
export enum SelectionMode {
  selection,
  deselection
}

// this will show message 'selection' as string representation of enum member
alert(SelectionMode[SelectionMode.selection]);
```

2. We could set the values equal to a string, ex:
```
export enum SelectionMode {
  selection = 'selection',
  deselection = 'deselection'
}

// this will show message 'selection' as string representation of enum member
alert(SelectionMode.selection);
```

I think the `enum` makes the front end implementation simpler and provides less room for error. Thought's on these approaches?

projects/go-lib/src/lib/components/go-table/go-table.component.ts
grahamhency: In this case, `row` is the data for the row in question. Right now, we're treating this as an `any[]`. See: `tableData: any[]`. We could _probably_ make the assumption they'll pass in an `object`, but I'm unsure. I think in this case, we could probably leave this as `any`. Thoughts on that?

projects/go-tester/src/app/app.service.ts
grahamhency: @AlexOverbeck this is just the test service.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
AlexOverbeck: Should `tableData` be `any` or is it an array of objects? Would it work if it was an array of something else? I'm more so curious than I am suggesting that we change it.

It looks to me like the function uses it like it is an object right? I guess row could technically be an array.

projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
AlexOverbeck: @grahamhency, I think option number 2 here seems like it makes the most sense to me. What do you think?

PR #133, opened 2019-08-15
Datepicker Component

### LINKED ISSUE #121: Create a <go-datepicker> component
We need to create a component that will allow users to select a date. We should probably have some discussion around what this UI looks like as it could take many forms.

#### Potential questions we should ask first:
1. Do we want to go all out with a calendar popup?
    _Accessibility and Internationalization are hard here._
2. Do we want to just have an input with some validations?
    _Could be hard to nail down an effective UX here._
3. Could we have three individual inputs for year, month, and day?
    _Certainly the easiest to control and maintain, but less cool than the first option._
4. What if we just used the native `<input type="date">`?
    _I'm not sure what the cross browser experience is here, but this one is the easiest to implement and control._

We should discuss our options in depth before proceeding with this one.

**Feature requirements include but are not limited to:**
2. Have a way for a user to input a date
3. Have a way to specify the format the date is saved in
4. Have a place to put a hint text to add additional information around the input
5. Handle its own errors with the addition of an error hint text.
6. Be able to be disabled
7. Support both a light and a dark theme
8. Be set up to work with Angular's [reactive forms](https://angular.io/guide/reactive-forms)

StevenUlmer:
closes #121 

Adds a datepicker component. This component relies mostly on angular materials datepicker with a few overrides to allow typing a date in with a locale. It's expecting to be used within reactive forms so it takes a form and group on the component. It doesn't allow you to pass a pattern in to match the date format to, but it allows you to pass a locale in and then it formats the date based on that locale. 

For example if the locale is `en-US` (English United States) then the format would be `mm/dd/yyyy` but if the locale is `de` (Germany) then the format would be `dd.mm.yyyy`.


StevenUlmer:
This has been completed by https://github.com/mobi/goponents/pull/151

PR #136, opened 2019-08-22
Manually updates classes when state changes to increase performance

AlexOverbeck:
This updates the accordion panel to use a simple object in the templates rather than functions. This forces us to manually update when the state changes but increases performance. 

Previous Angular was performing DOM manipulation changes when any event happened. Now it does not because it is more aware of when that happens.

### DIFF
--- a/projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.html
+++ b/projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.html
@@ -1,11 +1,11 @@
 <section
   class="go-accordion-panel"
-  [ngClass]="containerClasses()"
+  [ngClass]="containerClasses"
 >
   <header
     aria-expanded="expanded"
     class="go-accordion-panel__header"
-    [ngClass]="headerClasses()"
+    [ngClass]="headerClasses"
     (click)="toggle.emit()"
   >
     <div class="go-accordion-panel__title">
--- a/projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.html
+++ b/projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.html
@@ -3,7 +3,7 @@
   [ngClass]="containerClasses"
 >
   <header
-    aria-expanded="expanded"
+    [attr.aria-expanded]="_expanded"
     class="go-accordion-panel__header"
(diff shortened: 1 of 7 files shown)


projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.html
```diff
@@ -1,11 +1,11 @@
 <section
   class="go-accordion-panel"
-  [ngClass]="containerClasses()"
+  [ngClass]="containerClasses"
```
AlexOverbeck: This is the big win here^ Angular will no longer update these class attribute on any javascript event.

projects/go-lib/src/lib/components/go-accordion/go-accordion.component.ts
```diff
@@ -47,6 +47,7 @@ export class GoAccordionComponent implements OnInit, AfterContentInit {
       // NOTE: This feels a little destructive.
       // We lose track of the icon explicitly set by the child component.
       panel.icon = this.showIcons ? panel.icon : null;
+      panel.updateClasses();
```
AlexOverbeck: The minor trade off is that we have to manually trigger the updateClass function because this method of manually setting a child's attributes doesn't fire lifecycle events.

projects/go-lib/src/lib/components/go-accordion/go-accordion.component.ts
```diff
@@ -74,10 +75,12 @@ export class GoAccordionComponent implements OnInit, AfterContentInit {
 
   private openPanel(panel: GoAccordionPanelComponent): void {
     panel.expanded = true;
+    panel.updateClasses();
   }
 
   private closePanel(panel: GoAccordionPanelComponent): void {
     panel.expanded = false;
```
StevenUlmer: What if we added a setter to the expanded property that then updated the classes as well? This would move the responsibility of resetting the classes back to the child component.

projects/go-lib/src/lib/components/go-accordion/go-accordion.component.ts
StevenUlmer: if we added the setter to the expanded class we would have to move the expanded property to be the last thing updated in this method

PR #137, opened 2019-08-23
Add Storybook

eligolding:
This adds [Storybook](https://storybook.js.org/) to the project, with just a few stories to start with, as demoed at the last UI counsel meeting.


eligolding:
As per discussions, I updated this PR so that all stories are living with `/projects/stories`. This way we can merge this is in now so that people can start using and testing it to see if it fits our needs, and if we want to revert it we can do so in the least intrusive way.

@grahamhency Should I add some instructions to the readme, or better not since we don't know if we will be using this long term?

grahamhency:
Closing for now. We will revisit something like this in the future. Thanks @eligolding

Issue #115, opened 2019-08-05
Create <go-select> component


AlexOverbeck:
### What is the feature that you would like to propose?
We need to create a componentized version of a select box to make it easier to control implementation and features for the select element.

### Describe the solution you'd like
The functionality that should be included in the initial select box component are a bit of a moving target. 

**At the very least it should:**
1. Have a label that corresponds to the input
2. Have a select box that allows users to select either a single or multiple options
3. Have a place to put a hint text to add additional information around the input
4. Handle its own errors with the addition of an error hint text.
5. Be able to be disabled
6. Support both a light and a dark theme
7. Be set up to work with Angular's [reactive forms](https://angular.io/guide/reactive-forms)

PR #146, opened 2019-08-29
Go Select

devchris:
This closes https://github.com/mobi/goponents/issues/115

The go-select relies on ng-select as a peer dependency.


projects/go-lib/src/lib/components/go-select/go-select.component.ts
```diff
@@ -0,0 +1,18 @@
+import { Component, Input, ViewEncapsulation } from '@angular/core';
+import { FormControl } from '@angular/forms';
+
+@Component({
+  encapsulation: ViewEncapsulation.None,
```
grahamhency: Do we _have_ to do this? If so, what's the reasoning for it?

projects/go-lib/src/lib/components/go-select/go-select.component.ts
```diff
@@ -0,0 +1,18 @@
...
+  selector: 'go-select',
+  templateUrl: './go-select.component.html',
+  styleUrls: ['./go-select.component.scss']
+})
+export class GoSelectComponent {
+  @Input() bindLabel: string;
+  @Input() bindValue: string;
+  @Input() control: FormControl;
+  @Input() hints: string[];
+  @Input() items: Array<any>;
```
grahamhency: Should we switch this to `any[]` just for consistency in declaring our types?

projects/go-tester/src/app/app.component.scss
```diff
@@ -1 +1 @@
-
+@import "~@ng-select/ng-select/themes/default.theme.css";
```
grahamhency: I know this is for our tester app, but do we need to import this in the project separately? It looks like it's being imported in the component's SCSS file as well, are those styles not applying?

projects/go-lib/src/lib/components/go-select/go-select.component.html
```diff
@@ -0,0 +1,18 @@
+<ng-select
+  [items]="items"
+  bindLabel="{{bindLabel}}"
+  [multiple]="multiple"
+  [formControl]="control">
+</ng-select>
+
+<go-hint *ngFor="let hint of hints" [message]="hint" [theme]="theme"></go-hint>
+
+<div *ngIf="control?.errors?.length">
```
grahamhency: We probably don't need this div here, could we substitute it for an `ng-container` instead?

projects/go-lib/src/lib/components/go-select/go-select.component.scss
```diff
@@ -0,0 +1,33 @@
+@import '~@tangoe/gosheets/base/_variables';
+@import '~@tangoe/gosheets/base/_mixins';
+@import "~@ng-select/ng-select/themes/default.theme.css";
+
+.ng-select .ng-select-container {
+  height: 1.5rem + $column-gutter !important;
```
grahamhency: Oh man, what's up with this `!important`?

projects/go-tester/src/app/components/test-page-3/test-page-3.component.ts
```diff
@@ -53,6 +75,10 @@ export class TestPage3Component {
       if (input in this.form.controls) {
         this.form.controls[input].setErrors(errors);
       }
+
+      if (this.validationSelectControl.value.name !== 'Ron') {
+        this.validationSelectControl.setErrors(errorResponse['validationSelectControl']);
```
grahamhency: Again, I know this is tester, but this is not quite working as I would expect. It looks like it only shows the error on form submit. Shouldn't it show this error until it's valid?

projects/go-lib/package.json
```diff
@@ -9,11 +9,13 @@
   "description": "A library of UI components for Angular v7+.",
   "license": "MIT",
   "peerDependencies": {
+    "@ng-select/ng-select": "^2.20.5",
     "@tangoe/gosheets": "^1.0.0"
   },
   "devDependencies": {
     "@angular/common": "^7.0.0",
     "@angular/core": "^7.0.0",
+    "@ng-select/ng-select": "^2.20.5",
```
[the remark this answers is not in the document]
devchris: probably only in the peer dependency

projects/go-tester/src/app/app.component.scss
devchris: This was a leftover... removed it

projects/go-lib/src/lib/components/go-select/go-select.component.scss
devchris: Specificity issues from the ng-select...

projects/go-tester/src/app/components/test-page-3/test-page-3.component.ts
devchris: It depends how we want it.. Since this is only a tester, I don't think I'm gonna make this change

projects/go-lib/src/lib/components/go-select/go-select.component.ts
devchris: Sadly, yes. Otherwise we would have to import the ng select stylesheet to our styles.scss as well. This was a way around. I'm open for other suggestions

projects/go-lib/src/lib/components/go-select/go-select.component.html
devchris: I copied it from the go-input. I thought we could keep it for consistency

projects/go-tester/src/app/components/test-page-3/test-page-3.component.ts
AlexOverbeck: This would rely on us handling the validations on the frontend. I think we can do that for things like `required`, but I'm not sure if we want to make assumptions about validation things.

Issue #127, opened 2019-08-07
[Feature] Create a GoFileUpload Component


AlexOverbeck:
### What is the feature that you would like to propose?
We need to create a componentized version of a standard file upload input to make it easier to control implementation and features for the input element.

### Describe the solution you'd like
1. Have a label that corresponds to the input
2. Have a input with a type of file
3. Have a place to put a hint text to add additional information around the input
4. Handle its own errors with the addition of an error hint text.
5. Be able to be disabled
6. Support both a light and a dark theme
7. Be set up to work with Angular's [reactive forms](https://angular.io/guide/reactive-forms)

If we have time we could potentially add a lot of flourishes to this component. Maybe take a look at codepen for some inspiration:
https://codepen.io/search/pens?q=file%20upload&page=1&order=popularity&depth=everything

PR #149, opened 2019-08-30
Go File Upload

devchris:
This attempts to close https://github.com/mobi/goponents/issues/127


grahamhency:
@StevenUlmer we've already discussed this a bit, but can you look into what it would take for us to make the file upload process something like that InVision mockup?
Screen Shot 2019-10-03 at 9 41 13 AM

StevenUlmer:
@grahamhency I think this is ready for another look

projects/go-lib/src/lib/components/go-file-upload/go-dragon-drop.directive.ts
```diff
@@ -0,0 +1,35 @@
+import { Directive, EventEmitter, HostBinding, HostListener, Output } from '@angular/core';
+
+@Directive({
+  selector: '[goDragonDrop]'
+})
+export class DragonDropDirective {
+  @Output() fileDropped: EventEmitter<any> = new EventEmitter<any>();
+
+  @HostBinding('class.go-file-upload--active') active: boolean = false;
```
grahamhency: @StevenUlmer what if we extracted this out into it's own thing? We could pass in the `go-file-upload--active` class as an `@Input` here and then handle the `drop` event in the `file-upload` component as an event handler. 

We don't _have_ to in this PR, could always extract it out later. But that would make this drag/drop functionality reusable which would be nice. 

Thoughts?

projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.html
```diff
@@ -0,0 +1,53 @@
...
+  <div class="go-file-upload__container" *ngIf="state === 'selecting' && !isLoading">
+    <go-icon icon="publish" iconModifier="large"></go-icon>
+    Drag and Drop File or Click to Browse
+  </div>
+  <div class="go-file-upload__container go-file-upload__container--loading" *ngIf="isLoading">
+    <go-loader *ngIf="isLoading"></go-loader>
+    <span class="go-file-upload__container--loading-text">Uploading File...</span>
+  </div>
+  <div class="go-file-upload__container go-file-upload__container--selected" *ngIf="state === 'selected' && !isLoading">
+    <go-icon icon="check" iconModifier="positive" iconClass="go-icon--large"></go-icon>
```
grahamhency: We might want to check if @jaredami's PR is going to affect the `iconModifer` here. #281

projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.scss
```diff
@@ -0,0 +1,62 @@
+@import '../../../styles/variables';
+
+@mixin upload-active {
+  cursor: pointer;
+  background-color: $theme-light-bg-hover;
+  opacity: 0.8;
+}
+
+input[type="file"] {
```
grahamhency: Could we use a class for this instead?

projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.html
```diff
@@ -0,0 +1,53 @@
...
+  [ngClass]="{ 'go-file-upload--disabled': state === 'selected' || isLoading }"
+  (click)="filePicker.click()"
+  (fileDropped)="onFilePicked($event)"
+>
+  <div class="go-file-upload__container" *ngIf="state === 'selecting' && !isLoading">
+    <go-icon icon="publish" iconModifier="large"></go-icon>
+    Drag and Drop File or Click to Browse
+  </div>
+  <div class="go-file-upload__container go-file-upload__container--loading" *ngIf="isLoading">
+    <go-loader *ngIf="isLoading"></go-loader>
```
grahamhency: Can we change this to `loaderSize="small"`, I think it fits the space a little better.

projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.scss
```diff
@@ -0,0 +1,62 @@
...
+  visibility: hidden;
+}
+
+.go-file-upload {
+  height: 10rem;
+  border: 2px dashed grey;
```
grahamhency: Can we change this to `1px dashed $theme-light-border`? Fits a bit better with the rest of the styles.

projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.scss
```diff
@@ -0,0 +1,62 @@
...
+  border-radius: 5px;
```
grahamhency: Can we use the pre-defined border-radius of `$global-radius`

projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.ts
```diff
@@ -0,0 +1,72 @@
...
+  id: string;
+
+  @Input() control: FormControl;
+  @Input() hints: Array<string> = [];
+  @Input() isLoading: boolean = false;
+  @Input() key: string;
+  @Input() label: string;
+  @Input() multiple: boolean = false;
+  @Input() state: 'selecting' | 'selected' = 'selecting';
+  @Input() theme: 'light' | 'dark' = 'light';
```
grahamhency: This binding exists, but it doesn't seem this actually works as intended. Can you take a look at the styling for a `dark` theme and update them to be correct please?

projects/go-lib/src/public_api.ts
```diff
@@ -44,6 +44,10 @@ export * from './lib/components/go-datepicker/calendar-cell.model';
 export * from './lib/components/go-datepicker/go-datepicker.component';
 export * from './lib/components/go-datepicker/go-datepicker.module';
 
+// File Upload
+export * from './lib/components/go-file-upload/go-file-upload.component';
+export * from './lib/components/go-file-upload/go-file-upload.module';
+
```
grahamhency: If you decide to extract the DragonDrop out we'll want to make sure to export that as well.

projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.ts
grahamhency: We'll also want documentation for this.

projects/go-tester/src/app/components/test-page-3/test-page-3.component.html
```diff
@@ -45,6 +45,18 @@ <h2 class="go-heading-2">
                 ></go-input>
               </div>
 
+              <div class="go-column go-column--100">
+                <go-file-upload
+                  [control]="fileControl"
+                  buttonVariant="positive"
+                  buttonIcon="home"
```
grahamhency: It seems like maybe we don't need `buttonVariant` or `buttonIcon` here, correct?

projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.html
jaredami: I don't think my PR effects this. However, if I'm not mistaken, while the `go-button` takes "positive" as an option for its `buttonVariant` input (and converts it internally to "primary"), the `go-icon` only accepts "primary" for its `iconModifier` input, since it has no internal logic which converts "positive" to "primary" and only has a `go-icon--primary` class, and not a `go-icon--positive` class.

grahamhency: @jaredami you make a good point here. I think it might make more sense to set the color via a class from the `go-file-upload` component itself. Thought's @StevenUlmer ?

projects/go-lib/src/lib/components/go-file-upload/go-dragon-drop.directive.ts
StevenUlmer: I went ahead and pulled out al the functionality I could. I had to do some stuff around using class here because just using `@HostBinding('class')` overwrites all classes already applied to the element

projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.ts
StevenUlmer: I fixed the dark theming here and added documentation for it.

projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.html
StevenUlmer: Ya that makes sense. I added a class in the `go-file-upload` component to set the color

PR #150, opened 2019-09-03
Add sass-lint

### LINKED ISSUE #147: Create a sass-lint configuration file
We need to implement a sass linter configuration that matches up with the configuration defined in the gosheets repo.

#### gosheets configuration:
https://github.com/mobi/gosheets/blob/master/.sass-lint.yml

eligolding:
Rules copied from gosheets

Also added a pre-commit hook to prevent files from being commited without passing the linter.

fixes #147

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -1,0 +1,118 @@
+options:
+  formatter: stylish
+files:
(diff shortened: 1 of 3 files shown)


eligolding:
Removed pre-commit hook.

PR #167, opened 2019-09-09
Ability to Brand Side Nav and Accordion

### LINKED ISSUE #158: [Feature] Branding Service
We need a service that can be utilized by both the goponents library and by the applications that implement it. It should allow us to define a primary color that can then be used to apply a branded accent to various parts of the application.

_**Example:**_
> In the navigation component, we show a small vertical rectangle on the active route. This would need to be branded.

I am envisioning an Angular service that we can use to inject into any component necessary and then reference the brand color.

This feature will need to have the following requirements:
1. It should be able to be reused by any component, inside the library or out
2. It should be able to change and modify the UI without page reload
3. It should be integrated into the layout components
4. It should default to our `$base-primary` SASS variable
5. It needs to be performant

StevenUlmer:
closes #158

This adds the ability to brand a couple items in goponents. For now this just includes the 'active' rectangle for both the side nav and the accordion panels. This also removes some unused css from the go-card where branding was being used.

All it takes to brand everything is to import the GoBrandingService into your app.component.ts file and set the brand color through `goBrandingService.setBrandColor(color)`


grahamhency:
@StevenUlmer just a couple of other comments on this one.
1. There are some merge conflicts now
2. We will need to add some documentation around this. I am okay if you want to open a follow up PR for this after this is merged, but we'll want it included in `v1.3.0`

projects/go-lib/src/lib/components/go-side-nav/go-nav-item/go-nav-item.component.ts
```diff
@@ -1,14 +1,29 @@
-import { Component, Input } from '@angular/core';
+import { Component, Input, OnInit } from '@angular/core';
 import { NavItem } from '../nav-item.model';
 import { GoSideNavService } from '../go-side-nav/go-side-nav.service';
+import { GoConfigService } from '../../../go-config.service';
+import { GoConfigInterface } from '../../../go-config.model';
+import { distinctUntilKeyChanged } from 'rxjs/operators';
```
grahamhency: @StevenUlmer we can probably remove this line if we're not going to implement `distrinctUntilKeyChanged`

projects/go-lib/src/public_api.ts
```diff
@@ -18,6 +18,9 @@ export * from './lib/components/go-action-sheet/go-action-sheet.module';
 export * from './lib/components/go-badge/go-badge.component';
 export * from './lib/components/go-badge/go-badge.module';
 
+// Branding
+export * from './lib/go-config.service';
```
grahamhency: Should we also export the `GoConfigInterface` here?

projects/go-lib/src/lib/go-config.service.ts
```diff
@@ -0,0 +1,16 @@
...
+import { BehaviorSubject } from 'rxjs';
+import { GoConfigInterface } from './go-config.model';
+
+@Injectable()
+export class GoConfigService {
+  config: BehaviorSubject<GoConfigInterface> = new BehaviorSubject<GoConfigInterface> ({
+    brandColor: '#65B360'
+  });
+
+  public setBrandColor(color: string): void {
```
grahamhency: I like this idea of having utility methods on the service to set new values, 🅰️ ➕

projects/go-lib/src/lib/components/go-side-nav/go-nav-item/go-nav-item.component.ts
grahamhency: Actually, did you `pipe` it? 
```
this.configService.config.pipe(
  distinctUntilKeyChanged('brandColor')
).subscribe((value: GoConfigInterface) => {
  this.theme = value.brandColor;
});
```

StevenUlmer: I did. It ended up having something to do with how I was updating the config in the service. It works now

projects/go-lib/src/public_api.ts
StevenUlmer: Yes we should

Issue #159, opened 2019-09-06
[Chore] Remove Scale Effect Between Route Transitions


grahamhency:
### What general chore or task would you like to see performed?
The scale effect that happens on route changes is too abrasive of a change, we should remove it for now.

### Are there any specific steps you would like completed?
1. Update Angular animations for layout route transitions to remove scale effect

patryk0605:
I will be working on this issue

PR #173, opened 2019-09-10
Remove scale animation from route transitions

patryk0605:
https://github.com/mobi/goponents/issues/159


projects/go-lib/src/lib/animations/route.animation.ts
```diff
@@ -20,16 +18,7 @@ export const routerAnimation =
 
       query(':leave',
         [
-          style({ opacity: 1 }),
-          animate(timing + easing, style({ opacity: 0, transform: 'scale(0.95)' }))
```
AlexOverbeck: What was the reasoning for removing the transition here?

patryk0605:
@AlexOverbeck I figured linking the issue would be a good enough description as it defines purpose of the change. I can always copy what it's in the issue as not sure what more I can add.

projects/go-lib/src/lib/animations/route.animation.ts
patryk0605: ` style({ opacity: 1 })` is still apart of the `:leave` transition (just the diff is making it look weird). As for the animation, I didn't see a benefit leaving it in without the scale. I'll put them back in without the scale and see I could see a difference 😄

patryk0605: I've added back the animations

PR #176, opened 2019-09-11
Fix button icon alignment

### LINKED ISSUE #175: [BUG] go-icon styles are not applying correctly when nested inside go-button
**Describe the bug**
Because of view encapsulation, the styles we apply to icons inside of the go-button component are not applying correctly.

**To Reproduce**
1. Go to http://www.tangoe.design/ui-kit/button
2. Scroll down to where a button exists with an icon inside of it
4. Notice that the alignment is off

**Expected behavior**
I would expect the styles to be applied consistently in all instances. The icon should be aligned with the text.

**Screenshots**
button-issue

AlexOverbeck:
### Fixes issue with how the icon was aligned inside the button
View encapsulation was causing some of our button styles to not
be applied. This moves the styles out of the button component and
into the icon component so that they apply to the correct component.

### Clean up how classes are applied to both the go-button and go-icon
Previously, we used functions inside the template to declare the
classes on both the icon and button components. Switching this to
use lifecycle hooks and a simple object increases performance.

### Fixes potential issue with datepicker unsubscribe function
There are potential cases where the ngOnDestroy lifecycle hook
could be fired when a subscription doesn't exist. This causes the
unsubscribe function to fail and was throwing errors in our tests.
Checking for a subscription in those lifecycle hooks fixes the
issue.

### DIFF
--- a/projects/go-lib/src/lib/components/go-icon/go-icon.component.ts
+++ b/projects/go-lib/src/lib/components/go-icon/go-icon.component.ts
@@ -9,6 +9,7 @@ export class GoIconComponent {
   @Input() icon: string;
   @Input() iconModifier: string;
   @Input() iconClass: string;
+  @Input() disabled: boolean = false;
 
   public classObject(): object {
     let classes: object = {}; // tslint:disable-line:prefer-const
@@ -19,6 +20,10 @@ export class GoIconComponent {
 
     if (this.iconClass) {
       classes[this.iconClass] = true;
+    }
+
+    if (this.disabled) {
+      classes['go-icon--disabled'] = true;
     }
 
     return classes;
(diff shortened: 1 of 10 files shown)


projects/go-lib/src/lib/components/go-icon/go-icon.component.ts
```diff
@@ -1,26 +1,31 @@
...
   styleUrls: ['go-icon.component.scss']
 })
-export class GoIconComponent {
+export class GoIconComponent implements OnChanges {
+  classObject: object = {};
+
   @Input() icon: string;
   @Input() iconModifier: string;
   @Input() iconClass: string;
+  @Input() disabled: boolean = false;
```
StevenUlmer: Does it matter that disabled is an html attribute here? Should we change this to be iconDisabled or something instead?

AlexOverbeck: I think so long as it doesn't end up being `disabled="false"` we should be okay with it because it follows the html spec. I will make sure that isn't the case.

AlexOverbeck: I can confirm this. So long as `[disabled]` is an expression binding I think we should prefer this.

PR #177, opened 2019-09-11
Go Select Label

### LINKED ISSUE #174: [BUG] Go Select Label
**Describe the bug**
The go-select component doesn't contain an option for a label like the rest of the form components do.

**Expected behavior**
I would expect the go-select component to handle the label for me instead of having to implement it myself.

**Potential Solution**
Add a label Input to the go-select component that maps to a label in the html.

StevenUlmer:
This adds a label to the go-select component.

closes #174

### DIFF
--- a/projects/go-lib/src/lib/components/go-select/go-select.component.html
+++ b/projects/go-lib/src/lib/components/go-select/go-select.component.html
@@ -1,4 +1,12 @@
+<label
+  class="go-form__label"
+  [attr.for]="id"
+  [ngClass]="{'go-form__label--dark': theme === 'dark'}">
+  {{ label }}
+</label>
+
 <ng-select
+  [labelForId]="id"
   [items]="items"
   bindLabel="{{bindLabel}}"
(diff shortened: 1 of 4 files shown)


projects/go-lib/src/lib/components/go-select/go-select.component.html
```diff
@@ -1,4 +1,12 @@
+<label
+  class="go-form__label"
+  [attr.for]="id"
+  [ngClass]="{'go-form__label--dark': theme === 'dark'}">
+  {{ label }}
+</label>
+
 <ng-select
+  [id]="id"
```
AlexOverbeck: It looks like ng-select has a `[labelForId]` input that we could use to better associate the label to the input.

https://github.com/ng-select/ng-select/blob/master/src/ng-select/lib/ng-select.component.ts#L106

StevenUlmer: Good catch. Switched it to use [labelForId]

PR #209, opened 2019-09-26
Go-Footer

### LINKED ISSUE #164: [Feature] Layout Footer
The layout components need to be able to have a section to render a footer.

_**Example:**_
> https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d4dd6199dada53893b6a783

I think we could simply create a slot for content projection in the go-layout component. Style it slightly to justify-center and that would be sufficient.

### This feature will need to have the following requirements:
1. Included in the layout component
2. Center aligned in the app

StevenUlmer:
closes #164 

This creates a footer to be used within the layout. The footer isn't sticky and will appear at the end of the main content body after a user scrolls to the end. It uses content projection so the implementer can decide what they want to display inside of the footer. The only thing we're doing is centering the content that the user puts inside.


grahamhency:
@StevenUlmer after you fix the issue I mentioned before, we'll need to add documentation for this.

PR #210, opened 2019-09-27
Table Column Sorting

### LINKED ISSUE #180: [Feature] Table Sort Column Granularity
Currently, we can mark a table as sortable or not. This is a binary operator that either allows all columns to be sorted, or none of them. We need the ability to mark *specific* columns to be sortable. 

_**Example:**_
Our reporting framework currently provides this functionality. We need to mimic that. 

The ability to mark a subset of all table columns to be sortable on an individual basis.

StevenUlmer:
closes #180 

Currently we have the ability to either have every column on a table sortable or none of them. There
are certain cases where we want to be able to just sort by a couple columns on a table. This provides a way for an implementor to decide which columns can be sortable on a table or which columns they want to not be sortable. If a column doesn't contain the sortable binding it will default to whatever is in the table config.

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table-column.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table-column.component.ts
@@ -9,6 +9,7 @@ export class GoTableColumnComponent {
   @Input() field: string;
   @Input() title: string;
   @Input() width: number;
+  @Input() sortable?: boolean;
 
   @ContentChild('goTableCell') goTableCell: TemplateRef<any>;
   @ContentChild('goTableHead') goTableHead: TemplateRef<any>;
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.ts
@@ -1,4 +1,5 @@
 import {
+  AfterContentInit,
   Component,
   ContentChildren,
   ElementRef,
@@ -31,7 +32,7 @@ import { fadeTemplateAnimation } from '../../animations/fade.animation';
   templateUrl: './go-table.component.html',
   styleUrls: ['./go-table.component.scss']
 })
-export class GoTableComponent implements OnInit, OnChanges {
+export class GoTableComponent implements AfterContentInit, OnInit, OnChanges {
 
   @Input() loadingData: boolean = false;
   @Input() showTableActions: boolean = false;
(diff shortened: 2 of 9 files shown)


projects/go-lib/src/lib/components/go-table/go-table-column.component.ts
```diff
@@ -9,6 +9,7 @@ export class GoTableColumnComponent {
   @Input() field: string;
   @Input() title: string;
   @Input() width: number;
+  @Input() sortable?: boolean;
```
grahamhency: Should we just default this to true and then we don't have to do the `undefined` check?

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -103,14 +112,13 @@ export class GoTableComponent implements OnInit, OnChanges {
...
-    const { sortable, sortConfig, tableData }:
+    const { sortConfig, tableData }:
     {
-      sortable: boolean,
       sortConfig?: GoTableSortConfig,
       tableData: any[]
     } = this.localTableConfig;
 
-    if (tableData && sortable) {
+    if (tableData) {
```
grahamhency: I think we want to keep this because this will restrict sorting on a table level, not a column level.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -71,6 +72,14 @@ export class GoTableComponent implements OnInit, OnChanges {
     this.renderTable();
   }
 
+  ngAfterContentInit(): void {
```
grahamhency: If we default sorting to `true` on a column, then we won't have to worry about any of this.

grahamhency:
@StevenUlmer did you want to have a discussion about the comments I left on this? Also, we'll want to add documentation for this.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
StevenUlmer: @grahamhency We can default sorting to true, but not defaulting it allows us to have a little more flexibility in how the table is implemented. This way instead of just turning sorting off for a couple columns, we can turn off sorting for the entire table and then enable sorting on a couple columns. The only way to do this is to always keep track of sorting on the column instead of the table, but if this isn't something we care about theres a few things I can clean up here that wouldn't be necessary.

StevenUlmer:
@grahamhency I've removed that afterContentInit by changing some of the html. Can you take another look?

StevenUlmer:
@grahamhency Documentation has been written

grahamhency:
Merged, thanks for writing the docs!

PR #242, opened 2019-10-04
Switch toggle label position

### LINKED ISSUE #233: [Feature] Ability to specify which side to display label on go-switch-toggle
We would like the ability to specify a side to display the switch toggle label on.

A binding where we can pass `left` or `right`.

1. The ability to specify which side the label for the go-switch-toggle appears
2. It should default to appear on the right

eligolding:
fixes #233 

Following [material's slide-toggle](https://material.angular.io/components/slide-toggle/api), I named the binding `labelPosition` and set the options as `before` and `after`.


eligolding:
Wait, what? Something weird happened when switching base branch. gonna close and open new one

PR #260, opened 2019-10-08
Add appendTo input to go-select

### LINKED ISSUE #246: [BUG] Clicking on the dropdown of a go-select within a go-modal closes the modal
**Describe the bug**
Making a selection on a dropdown of a `go-select` within a `go-modal` causes the modal to close.

**Expected behavior**
Clicking a `go-select` drop-down item within a `go-modal` should not automatically close the modal.

jaredami:
Closes #246 

Currently, making a selection on a `go-select` within a `go-modal` causes the modal to close. Sometimes we might desire that the `go-modal` remain open until a submission button is clicked, for instance.

By adding an `appendTo` input to our `go-select` and passing it to the `appendTo` input of the contained `ng-select`, we can make it so that a `go-modal` will remain open upon making a selection to a `go-select` within it using `appendTo="'body'"`.

Example:
```
openModal(): void {
    this.goModalService.openModal(
      GoSelectComponent,
      {
        appendTo: 'body',
        bindLabel: 'name',
        bindValue: 'value',
        control: this.select10,
        items: this.items,
        label: 'Favorite Candy',
      }
    );
  }
```

### DIFF
--- a/projects/go-lib/src/lib/components/go-select/go-select.component.ts
+++ b/projects/go-lib/src/lib/components/go-select/go-select.component.ts
@@ -10,6 +10,7 @@ import { FormControl } from '@angular/forms';
 export class GoSelectComponent implements OnInit {
   id: string;
 
+  @Input() appendTo: string;
   @Input() bindLabel: string;
   @Input() bindValue: string;
   @Input() control: FormControl;
--- a/projects/go-lib/src/lib/components/go-select/go-select.component.html
+++ b/projects/go-lib/src/lib/components/go-select/go-select.component.html
@@ -6,6 +6,7 @@
 </label>
 
 <ng-select
+  [appendTo]="appendTo"
   [labelForId]="id"
   [items]="items"
   [bindLabel]="bindLabel"
--- a/projects/go-style-guide/src/app/features/ui-kit/components/form-docs/components/select-docs/select-docs.component.html
+++ b/projects/go-style-guide/src/app/features/ui-kit/components/form-docs/components/select-docs/select-docs.component.html
@@ -318,4 +318,30 @@
       </div>
     </ng-container>
   </go-card>
+  <go-card id="form-select-appendto" class="go-column go-column--100">
+    <ng-container go-card-header>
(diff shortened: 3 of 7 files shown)


projects/go-lib/src/lib/components/go-select/go-select.component.ts
```diff
@@ -10,6 +10,7 @@ import { FormControl } from '@angular/forms';
 export class GoSelectComponent implements OnInit {
   id: string;
 
+  @Input() appendTo: string;
```
grahamhency: @jaredami do we need a default here?

grahamhency: How does this work if we pass undefined to `ng-select`?

jaredami: Good question. From [the docs:](https://ng-select.github.io/ng-select#/append-to-element) "By default dropdown panel is rendered as child element of ng-select, but you can append dropdown to any element using appendTo." 

If we were to set a default value, I'm not sure what it should be so that it would match the default behavior.

At any rate, I tested it with passing nothing in as well explicitly passing in `undefined`, and in both cases it appears to behave the same as the default. So it seems to work fine without explicitly naming a default. Am I missing something?

grahamhency: @jaredami nah, sounds good to me.

eligolding:
@jaredami Sorry for commenting on something so old, but just wanted to bring up that this solution introduces other issues. I am experiencing both of these issues that others have had with ng-select: https://github.com/ng-select/ng-select/issues/1130 and https://github.com/ng-select/ng-select/issues/929. The real solution will need to involve fixing the modal not to close, which @patryk0605 may have fixed here: https://github.com/mobi/goponents/pull/410

jaredami:
@eligolding Gotcha. Thanks for filling me in 👍

IoSono000:
Hi, i continue to have the same problem. If I insert the component in a simple html page no problem, but if I insert it in a more complex context, the problem remains ...

Issue #197, opened 2019-09-22
[Chore] Update Box Shadows


grahamhency:
### What general chore or task would you like to see performed?
Box shadows need to be updated to reflect design spec. 
https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d7a967497799142a093ebd2

### Are there any specific steps you would like completed?
- [ ] Update and add new SCSS variables
  - [ ] Update `$global-box-shadow` to the "Normal Box Shadow" in the DSM
  - [ ] Add variable for "High Box Shadow"
  - [ ] Add variable for "Search Box Shadow"
  - [ ] Add variable for "Shell Dropdown Box Shadow"
- [ ] Update components using current shadow that should be using alternate shadow.
  - [ ] `go-action-sheet` should use the "Shell Dropdown Box Shadow"
  - [ ] `go-search` should use the "Search Box Shadow"
  - [ ] `go-header` should use the "High Box Shadow"

grahamhency:
Waiting on UX sync for clarification on usage of shadows, which components get which shadows

grahamhency:
This task has all of the requirements. We're good to go here.

patryk0605:
I can take this issue

PR #261, opened 2019-10-09
Update box-shadows

patryk0605:
## Why do we need this?

This PR is required to address the box shadow requirements specified within the [DSM](https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d7a967497799142a093ebd2)

## How does this solve that?

Uses defined box-shadows. 

## Additional Info
Closes https://github.com/mobi/goponents/issues/197


patryk0605:
Pending better variable names from Kim (to be more generic).

grahamhency:
@patryk0605 to not hold up this PR any further, let's keep the search component using the old box-shadow as it provides better separation with its background. We can follow up with UX and determine if we need to update the shadow in a future release.

PR #271, opened 2019-10-14
[Chore] Button Design Update

### LINKED ISSUE #170: [Chore] Update Button Styles
Since the original design of the go-button, we've updated the styling and variations. We need to update our go-button to reflect the design spec.

1. Add variation options for each type outlined in DSM
2. Ensure each state is accounted for (hover, selected, disabled, etc. Check DSM for all states)

grahamhency:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

We have button variants for:
- `positive`
- `negative`
- `neutral`

Issue Number: Closes #170 

## What is the new behavior?
We have button variants for:
- primary (replacing positive)
- `secondary`
- `tertiary`
- `neutral`
- `negative`

- [x] Yes
- [ ] No

There are two changes that could be considered "breaking".
1. The `alert` variant is being removed. This was already deprecated in a previous version, but is officially being removed now.
  - Any instances of `alert` that remain should be changed to `negative`
2. The `positive` variant is being removed, however, we are going to accept it as a variant and change it to `primary` for now. In this case, it's not functionality breaking, but possibly UX breaking.
  - Not all `positive` buttons will need to be `primary`. In that case, the buttons labeled as `positive` should be evaluated on a per-component basis to determine if they need to be `primary` or instead should be `secondary` or `tertiary`.

## Other information

### DIFF
--- a/projects/go-lib/src/lib/components/go-button/go-button.component.ts
+++ b/projects/go-lib/src/lib/components/go-button/go-button.component.ts
@@ -64,7 +64,7 @@ export class GoButtonComponent implements OnChanges, OnInit {
       this.loaderClassObject['go-button__loader--light'] = true;
     } else {
       this.loaderType = 'light';
-      this.loaderClassObject['go-button__loader--' + this.buttonVariant] = true;
+      this.loaderClassObject[`go-button__loader--${this.buttonVariant}`] = true;
     }
   }
 
--- a/projects/go-lib/src/lib/components/go-button/go-button.component.ts
+++ b/projects/go-lib/src/lib/components/go-button/go-button.component.ts
@@ -1,20 +1,30 @@
-import { Component, EventEmitter, Input, Output, OnChanges } from '@angular/core';
+import {
+  Component,
+  EventEmitter,
+  Input,
+  OnChanges,
+  OnInit,
+  Output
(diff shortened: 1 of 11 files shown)


projects/go-lib/src/lib/components/go-button/go-button.component.ts
```diff
@@ -25,20 +35,44 @@ export class GoButtonComponent implements OnChanges {
...
     this.classObject = {
       'go-button--dark': this.useDarkTheme,
-      'go-button--loading': this.isProcessing,
-      'go-button--negative': isNegative,
-      'go-button--neutral': (this.buttonVariant === 'neutral'),
-      'go-button--positive': (this.buttonVariant === 'positive')
+      'go-button--loading': this.isProcessing
     };
+
+    this.classObject['go-button--' + this.buttonVariant] = true;
```
mwagz: Should we use string interpolation here?

```typescript
this.classObject[`go-button--${this.buttonVariant}`] = true;
```

grahamhency: 🤷‍♂ Sure.

PR #281, opened 2019-10-18
Add ability to brand top left header

### LINKED ISSUE #245: [Feature] Ability to brand the top left header area
We would like the ability to apply a branded background to the top left section of the header.

_**Example:**_
https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d5ea6e5cb739d2f806830bc

We should utilize the branding service to apply a 'branded' state to the top left section of the go-header. 

- [ ] Binding to allow branding for `go-header`
- [ ] Apply a color to the background of the menu/logo area in `go-header` if `go-header` is configured for `branding`
- [ ] Default behavior should be current implementation. (white background)
- [ ] When branding is applied, we **must** meet accessibility standards
  - [ ] This means we have to compare the background color with the color of the text and it must be ` >= 4.5`. [To calculate this in js.](https://stackoverflow.com/a/9733420)
  - [ ] We will also need to be able to provide a hex code, but the contrast ratio calculation is done with rgb. [To convert hex to rgb in js.](https://stackoverflow.com/a/5624139)
  - [ ] If the brand color provided is `< 4.5` contrast with the text color then we should use the opposite color for text. Those colors are `$theme-light-color` (default on white) and `$theme-dark-color`.
  - [ ] Ideally we could provide this as a function in the branding service for use elsewhere

jaredami:
Closes #245  

This PR adds the ability to apply a branded background to the top left section of the header.

Our `GoConfigService` already provides the ability to set a branding color. This PR adds the ability to toggle on and off whether this branding is applied to the header via the service's `toggleHeaderBrandingEnabled` method.

The header will use the service to ensure that the contrast ratio between the branded background and the color of the menu icon remain accessible by utilizing its `contrastIsAccessible` method.

### DIFF
--- a/projects/go-lib/src/lib/components/go-icon/go-icon.component.scss
+++ b/projects/go-lib/src/lib/components/go-icon/go-icon.component.scss
@@ -32,10 +32,10 @@ $icon-sizes: (small: 1rem, medium: 1.5rem, large: 2rem);
 }
 
 .go-icon--dark {
-  color: $theme-light-color;
+  color: $theme-dark-color;
 
   &.go-icon--disabled {
-    color: rgba($theme-light-color, .7);
+    color: rgba($theme-dark-color, .7);
   }
 }
 
@@ -43,10 +43,10 @@ $icon-sizes: (small: 1rem, medium: 1.5rem, large: 2rem);
 .go-icon--primary,
 .go-icon--negative,
 .go-icon--neutral {
-  color: $theme-dark-color;
+  color: $theme-light-color;
 
   &.go-icon--disabled {
-    color: rgba($theme-dark-color, .7);
+    color: rgba($theme-light-color, .7);
   }
 }
 
--- a/projects/go-lib/src/lib/components/go-icon/go-icon.component.scss
+++ b/projects/go-lib/src/lib/components/go-icon/go-icon.component.scss
@@ -32,10 +32,10 @@ $icon-sizes: (small: 1rem, medium: 1.5rem, large: 2rem);
 }
 
 .go-icon--dark {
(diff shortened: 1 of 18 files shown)


projects/go-lib/src/lib/go-config.service.ts
```diff
@@ -14,4 +21,47 @@ export class GoConfigService {
...
+    this.config.next(config);
+  }
+
+  public contrastIsAccessible(backgroundHex: string, foregroundHex: string): boolean {
+    const backgroundRgb: RGB = this.hexToRgb(backgroundHex);
+    const foregroundRgb: RGB = this.hexToRgb(foregroundHex);
+
+    const contrast: number = this.contrast(backgroundRgb, foregroundRgb);
+
+    return contrast > 4.5;
```
grahamhency: Technically, this can be `>=` to 4.5

jaredami:
@grahamhency I added tests for the GoConfigService and fixed the issue with the buttons you mentioned.

projects/go-lib/src/lib/components/go-icon/go-icon.component.scss
```diff
@@ -32,21 +32,21 @@ $icon-sizes: (small: 1rem, medium: 1.5rem, large: 2rem);
 }
 
 .go-icon--dark {
-  color: $theme-dark-color;
+  color: $theme-light-color;
```
grahamhency: @jaredami I think instead of changing this. We can just remove the `[iconModifier]` being passed to the `go-icon` by the `go-button`. See line 18 in `go-button.component.html`. The inheritance should take over here.

Issue #227, opened 2019-09-16
[Chore] CheckboxGroup Documentation


grahamhency:
We need documentation for the GoCheckboxGroupComponent

PR #283, opened 2019-10-22
[Chore] Checkbox Documentation

### LINKED ISSUE #228: [Chore] Checkbox Documentation
We need documentation for the GoCheckboxComponent.

grahamhency:
## Problem
We did not have documentation covering the `go-checkbox` or the `go-checkbox-group` components.

## Solution
Added documentation for both components.

Closes #228, #227

### DIFF
--- a/projects/go-style-guide/src/app/features/ui-kit/components/button-docs/button-docs.component.html
+++ b/projects/go-style-guide/src/app/features/ui-kit/components/button-docs/button-docs.component.html
@@ -107,7 +107,7 @@
           <go-button (handleClick)="testClick()" buttonDisabled="true">Disabled</go-button>
         </div>
         <div class="go-column go-column--33">
-          <go-button (handleClick)="testClick()" buttonIcon="work">With Icon</go-button>
+          <go-button (handleClick)="testClick()" buttonIcon="work"></go-button>
         </div>
       </div>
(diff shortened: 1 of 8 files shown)


projects/go-style-guide/src/app/features/ui-kit/components/button-docs/button-docs.component.html
```diff
@@ -107,7 +107,7 @@ <h2 class="go-heading-5">Primary Button</h2>
           <go-button (handleClick)="testClick()" buttonDisabled="true">Disabled</go-button>
         </div>
         <div class="go-column go-column--33">
-          <go-button (handleClick)="testClick()" buttonIcon="work">With Icon</go-button>
+          <go-button (handleClick)="testClick()" buttonIcon="work"></go-button>
```
StevenUlmer: shouldn't a `go-button` always have text? If it's just an icon button shouldn't it be a `go-icon-button` instead?

grahamhency: @StevenUlmer ... yeah. Jared and I talked about this. We're keeping it this way because the designs are fundamentally different between the two. Yet, UX is still showing this style of button with only an icon in it. So for now, 🤷‍♂

PR #286, opened 2019-10-25
Table Row Details

### LINKED ISSUE #250: [Feature] GoTable Detail Row
The ability to provide further details about a row of data in the table through a clickable arrow on the left most section of the row that expands a container element below said row. This container would be custom templated per the implementer.

_**Example:**_
The Kendo framework has a great working example of this:
https://www.telerik.com/kendo-angular-ui/components/grid/advanced-features/detail-template/

#### This feature will needs to have the following requirements:
- [x] Ability to render a row of detailed data associated with a parent row
- [x] It should be collapsable via a clickable arrow in the left most column of the parent rows
  - [x] By default, it will be collapsed
  - [x] By default, the arrow will point right
  - [x] When the row is expanded, the arrow should point down
- [ ] When the arrow is clicked, the table should emit a custom event to include:
  - [ ] The data of the row
  - [ ] Which direction the action took, expand/collapse
- [x] The detail row should be declared and formatted when building the table HTML (similar to the ability to format table cells and headers)
##### Extras:
- Provide animations for expanding/collapsing
- Provide a loader overlay for the detail row (used when making ajax requests after row is expanded)

StevenUlmer:
This adds the ability to use details within a table row. The implementor can provide a template
which is then used on every row to show the details when toggled on.

Fixes #250

### DIFF
--- a/projects/go-lib/src/lib/animations/table-details.animation.ts
+++ b/projects/go-lib/src/lib/animations/table-details.animation.ts
@@ -21,14 +21,14 @@ export const detailButtonAnim: AnimationTriggerMetadata = trigger('detailButtonA
   ])
 ]);
 
-export const detailsAnimation: AnimationTriggerMetadata = trigger('detailsAnimation', [
+export const tableRowBorderAnim: AnimationTriggerMetadata  = trigger('tableRowBorderAnim', [
   state('open', style({
-    display: 'table-row'
+    'border-bottom-color': 'rgba(240, 240, 240, 0)'
   })),
   state('close', style({
-    display: 'none'
+    'border-bottom-color': 'rgba(240, 240, 240, 1)'
   })),
   transition('open <=> close', [
-    animate('.125s ' + easing)
+    animate(timing + easing)
   ])
 ]);
--- a/projects/go-lib/src/lib/animations/table-details.animation.ts
+++ b/projects/go-lib/src/lib/animations/table-details.animation.ts
@@ -1,0 +1,34 @@
+import {
+  animate,
+  AnimationTriggerMetadata,
+  state,
+  style,
+  transition,
+  trigger
+} from '@angular/animations';
+
+import { easing, timing } from './_configs';
+
+export const detailButtonAnim: AnimationTriggerMetadata = trigger('detailButtonAnim', [
+  state('open', style({
+    transform: 'rotate(90deg)'
+  })),
+  state('close', style({
+    transform: 'rotate(0)'
+  })),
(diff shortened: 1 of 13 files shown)


projects/go-lib/src/lib/animations/table-details.animation.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const detailsAnimation: AnimationTriggerMetadata = trigger('detailsAnimation', [
+  state('open', style({
+    display: 'table-row'
+  })),
+  state('close', style({
+    display: 'none'
+  })),
+  transition('open <=> close', [
+    animate('.125s ' + easing)
```
grahamhency: Do you feel like we need to have a different timing for this?

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -33,27 +37,46 @@ <h5 class="go-heading-5 go-heading--no-margin" *ngIf="tableTitle">{{ tableTitle
...
-            </ng-template>
-          </td>
-        </tr>
+        <ng-container *ngFor="let item of setDisplayData()">
+          <tr class="go-table__body-row">
+            <td class="go-table__body-column"
+                [ngClass]="{ 'go-table__body-row--expanded': item.showDetails }"
+                width="40"
+                *ngIf="details">
+              <go-icon-button
```
grahamhency: So looking at this in the UI, the padding on the `icon-button` is making the row heights much taller than they probably should be. Do you think you could try to find a way around this?

projects/go-lib/src/lib/animations/table-details.animation.ts
grahamhency: Also, you can't perform an animation on `display` so this doesn't really do much visually.

projects/go-lib/src/lib/components/go-table/go-table.component.scss
```diff
@@ -213,3 +213,11 @@
     text-align: center;
   }
 }
+
+.go-table__body-row--expanded {
+  border-bottom: none;
```
grahamhency: What do you think about changing this to change the opacity of the color of the bottom border? That way we can animate the color and basically fade out the border.

projects/go-lib/src/lib/components/go-table/go-table.component.html
StevenUlmer: I think I fixed it by adding a class that offsets the padding of the row by the amount that's on the `icon-button`

PR #288, opened 2019-10-28
Update selected state background of table rows

### LINKED ISSUE #238: [Chore] Update table row selected state background
The table selection state background color needs updated to the design spec.

###### Example 
https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d82360a6beb090bbc3d6ba0

1. When a row is selected, `$theme-light-app-bg` should be applied as the background for that row

jaredami:
Closes #238 

We need to make it so that the background of a table row changes when it is selected.


grahamhency:
@jaredami, you'll want to rebase this after the base branch change. Also I am checking with UX on this because the selected state is the same as the application background and it tends to mess up the structure of the table..

PR #291, opened 2019-10-29
Update checkbox design

### LINKED ISSUE #235: [Chore] Update Checkbox Design
The design of our checkboxes need to be updated to meet the design spec. 

###### Example 
https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d3b2fd34b5f2e3140d29587

1. Each version of the checkbox should be represented in our component

jaredami:
Closes #235 

We need to update the design of our `go-checkbox` to meet the design spec, seen here: https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d3b2fd34b5f2e3140d29587

### DIFF
--- a/projects/go-lib/src/styles/_forms.scss
+++ b/projects/go-lib/src/styles/_forms.scss
@@ -113,6 +113,7 @@ $input--disabled-background: rgba($base-dark-secondary, .4);
   }
 }
 
+/* Hide the default checkbox */
 .go-form__checkbox--hidden {
   cursor: pointer;
   height: 0;
@@ -122,7 +123,7 @@ $input--disabled-background: rgba($base-dark-secondary, .4);
 }
 
 /* Create a custom checkbox */
-.go-form__checkbox-display {
+.go-form__custom-checkbox {
   background-color: $base-light;
   border: 1px solid lighten($base-light-secondary, 13.5%);
   border-radius: 2px;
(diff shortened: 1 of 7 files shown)


projects/go-lib/src/styles/_forms.scss
```diff
@@ -106,6 +113,66 @@ $input--disabled-background: rgba($base-dark-secondary, .4);
...
+
+/* When the checkbox is disabled and checked, change the background and border color of the custom checkbox */
+.go-form__checkbox:disabled:checked ~ .go-form__custom-checkbox {
+  background-color: lighten($base-light-secondary, 13.5%);
+  border: 1px solid lighten($base-light-secondary, 13.5%);
+}
+
+/* Create and style a custom checkmark (hidden when not checked) */
+.go-form__custom-checkbox:after {
+  border: solid #FDFDFD;
```
jaredami: I believe we need to create a variable for this, unless it is equivalent to one of our calculated colors.

projects/go-lib/src/styles/_forms.scss
```diff
@@ -106,6 +113,66 @@ $input--disabled-background: rgba($base-dark-secondary, .4);
...
+  opacity: 0;
+  position: absolute;
+  width: 0;
+}
+
+/* Create a custom checkbox */
+.go-form__custom-checkbox {
+  background-color: $base-light;
+  border: 1px solid lighten($base-light-secondary, 13.5%);
+  border-radius: 2px;
```
grahamhency: I think we can push this to 3px with our `global-radius` variable. Thoughts?

projects/go-lib/src/styles/_forms.scss
grahamhency: I think this should just be `$base-light`. If you look at the DSM it says `or Shuttle` and `Shuttle` is just `#fff` which is `$base-light`.

projects/go-lib/src/styles/_forms.scss
jaredami: It looks noticeably more rounded than the design spec that way, but it still looks fine in my opinion. 

So I guess it comes down to how important it is that we match the design spec as closely as possible vs how much we want to keep things consistent, e.g. by using our global variables 🤷‍♂

grahamhency: I would say that we should strive for consistency. Let's use the variable.

PR #293, opened 2019-10-30
Control state of per page dropdown from table config

### LINKED ISSUE #292: [BUG] Table per page dropdown always defaults to first option even when set
**Describe the bug**
When setting the `pageConfig` on a go-table, if you set the `perPage` to something other than the first option in the `pageSizes` array, the dropdown will still have the first option selected, even though the table will have the configured page count.

**To Reproduce**
1. Set `perPage` to something other than default:
```
  // ...stuffs
  pageConfig: {
    offset: 0,
    pageSizes: [25, 50, 100],
    perPage: 50, // second option
  },
});
```
2. Load the page in you browser de jour
3. See the unexpected results
Tangoe 2019-10-30 11-42-39

eligolding:
Sets the value of the dropdown based on `localTableConfig`

fixes #292

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.html
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.html
@@ -59,7 +59,7 @@
   </div>
   <div class="go-table-toolbar" *ngIf="showPaging()">
     <div class="go-table-toolbar__page-size">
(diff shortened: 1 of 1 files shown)


StevenUlmer:
Looks like the CI build passed, but it's not updating the status here. Gonna go ahead and merge it.

PR #309, opened 2019-11-07
File Upload Files Undefined

StevenUlmer:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/mobi/goponents/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?
Issue Number: https://github.com/mobi/goponents/issues/308

## What is the new behavior?
Depending on how we're getting the files (dragging and dropping vs clicking on the upload) the events from the html come with different attributes. This fixes the attribute properties for clicking on the upload

Fixes #308 

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #306, opened 2019-11-06
Pause Toaster On Hover

StevenUlmer:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/mobi/goponents/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?
Issue Number: https://github.com/mobi/goponents/issues/295

## What is the new behavior?
When a user hovers over a toast, the toast will cancel the timer until the user hovers off of the toast and then the timer is restarted

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

Issue #298, opened 2019-11-04
[Feature] Preselect All Options in Table


StevenUlmer:
### What is the feature that you would like to propose?
There are certain situations where we use a table to provide details about items that we are looking to update. In these situations the user has already searched for the items that they want to change and then we give them the table as a way of deselecting values that they might not actually want to update. In this case it would be nice to be able to auto select all options when the page loads instead of forcing the user to search for the items and then press select all.

### Describe the solution you'd like
Add an option to the table config to preselect all options. `preselected: false | true = false`. The rows in the table will be preselected or not based on that value

PR #315, opened 2019-11-12
v1.3.1

grahamhency:
###### `v1.3.1`

# 🚢 New Features

### Table
- Added ability to preselect all of the options in the table #298 
- Added ability to provide a detail row #286 

# 🐞 Bug Fixes
### File Upload
- Fixed an issue where if a user attempts to click on "Upload Files" and an error would be thrown in the console. #309 

### Layout
- Fixed an issue where the navigation active indicator in the side nav would not show up properly if the side nav was rendered inside of another `router-outlet`. #316 
- Fixed an issue where the navigation active indicator would not appear if the url contains params. #317 

# ✏️ General & Design

### General
- Toaster now pauses on hover #306 
- Updated karma version for tests #307 

### Design
- Table
  - Added row striping #237
  - Added a selected state color to entire rows #238


projects/go-lib/src/lib/components/go-file-upload/go-file-upload.component.spec.ts
```diff
@@ -110,4 +110,40 @@ describe('GoFileUploadComponent', () => {
...
+          files: files
+        }
+      });
+
+      expect(component.files.length).toBe(1);
+      expect(component.filePreview.length).toBe(1);
+      expect(component.filePreview[0]).toBe('file');
+    });
+
+    it('doesn\t error when clicking on the file upload', () => {
```
StevenUlmer: I feel like this is supposed to be `doesn\'t`

Issue #236, opened 2019-10-03
[Feature] Table Page Counts


grahamhency:
### What is the feature that you would like to propose?
We need to be able to show the page numbers up to 5 at a time.

_**Example:**_
https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d812544d56beb2dae5da0ec

### Describe the solution you'd like
The table needs to show up to 5 numbers at a time. 

This feature will needs to have the following requirements:
1. Support both Client and Server side functionality
2. Show a total number of up to 5 pages
3. If not on one of the first two or last two pages, then the active page should be the center page with two pages showing on each side.
4. The active page should be highlighted per the design
5. Hovering over other page numbers besides active should show a hover state per the design

grahamhency:
Closed with #318

PR #318, opened 2019-11-12
Add Page Numbers to Pagination

StevenUlmer:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/mobi/goponents/issues/236

## What is the new behavior?
On the go-table if there is pagination there will be 5 pages that show up at a time in the page numbers. They will be centered around the active page unless that page is one of the first two pages or last two pages and then it will just have the first or last 5 pages showing.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.ts
@@ -29,6 +29,7 @@ import {
 import { sortBy } from './go-table-utils';
 import { fadeTemplateAnimation } from '../../animations/fade.animation';
 import { detailButtonAnim, tableRowBorderAnim } from '../../animations/table-details.animation';
+import { GoTablePage } from './go-table-page.model';
 
 @Component({
   animations: [
@@ -68,7 +69,7 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
   selectAllChecked: boolean = false;
   targetedRows: any[] = [];
   showTable: boolean = false;
-  pages: any[] = [];
+  pages: GoTablePage[] = [];
 
   constructor(private changeDetector: ChangeDetectorRef) { }
 
@@ -97,7 +98,7 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
 
       this.setTotalCount();
       this.handleSort();
-      this.setPage(0);
+      this.setPage();
     }
 
     this.showTable = Boolean(this.tableConfig);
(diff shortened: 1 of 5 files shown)


projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -68,6 +68,7 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
   selectAllChecked: boolean = false;
   targetedRows: any[] = [];
   showTable: boolean = false;
+  pages: any[] = [];
```
grahamhency: Should we type this more specifically?

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -148,20 +150,24 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
...
+  isLastPage(offset: number): boolean {
     const { pageConfig, tableData, totalCount }:
     {
       pageConfig: GoTablePageConfig,
       tableData: any[],
       totalCount: number
     } = this.localTableConfig;
 
-    return ((pageConfig.offset + pageConfig.perPage) >= tableData.length) && ((pageConfig.offset + pageConfig.perPage) >= totalCount);
+    if (offset === undefined) {
```
grahamhency: Would `offset` be `undefined` if it's a required param? I could see if you wanted to check for this if it was optional and then doing something with it, but not sure this is needed.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -348,5 +358,50 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
...
+    } = this.localTableConfig;
+
+    const lastPage: number = Math.ceil(totalCount / pageConfig.perPage);
+    const currentPage: number = (pageConfig.perPage + offset) / pageConfig.perPage;
+    const startPage: number = this.calculateStartPage(lastPage, currentPage);
+
+    this.pages = [];
+
+    for (let i: number = startPage; i < startPage + 5; i++) {
+      if (i > lastPage) {
```
grahamhency: We could clean this up a little bit with an inline if here.
```
if (i > lastPage) break;

this.pages.push({
          number: i,
          active: i === currentPage
        });
```

projects/go-lib/src/lib/components/go-table/go-table.component.ts
grahamhency: Which, it might be easier to set a default for `offset`?
```
isLastPage(offset: number = 0): boolean {

}
```
Then we don't have to pass `0` in all the method calls in the file and it will never be `undefined`. Unless I am missing something.

StevenUlmer: I think you're right here. I think we want it to be a required param

PR #319, opened 2019-11-12
Prevent Accordion Content from Initializing Before Opened

StevenUlmer:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Accordions will always load anything inside of it into the dom when it's loaded. This causes content that may never be rendered on the page to be loaded when the page is initialized, which is inefficient. Ideally we would be able to not load this content until the accordion is actually opened.

## What is the new behavior?
There is an option depending on how you initialize the accordion to not load the content of an accordion panel until it's clicked on. This is done through the use of ng-template and will require an extra line when writing the html for each accordion pane. 

- [ ] Yes
- [x] No

There are no breaking changes, but going forward new accordion panels should probably be written with the ng-template instead of relying on ng-content because of the performance gains that this creates.

### DIFF
--- a/projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.ts
+++ b/projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.ts
@@ -6,7 +6,8 @@ import {
   Input,
   OnChanges,
   OnInit,
-  Output
+  Output,
+  TemplateRef
 } from '@angular/core';
 
 import { accordionAnimation } from '../../animations/accordion.animation';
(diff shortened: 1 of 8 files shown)


projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.ts
```diff
@@ -40,13 +42,18 @@ export class GoAccordionPanelComponent implements OnInit, OnChanges {
     this._expanded = expanded;
     this.containerClasses['go-accordion-panel--active'] = expanded;
     this.headerClasses['go-accordion-panel__header--active'] = expanded;
+    if (expanded) {
```
grahamhency: I wonder if we should have some sort of `persistState` property that you could configure to refresh the component every time or not and default it to persist. Thought's on that?

PR #321, opened 2019-11-14
[Feature] Support title and sizes for modal 

patryk0605:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/mobi/goponents/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Currently the modal does not support the ability for a title to be on the same line as `x` and does not support fixed size requirements per UX design.

Issue Number: https://github.com/mobi/goponents/issues/313

## What is the new behavior?

Two new binding have been added to modal: `modalTitle` and `modalSize`

## Does this PR introduce a breaking change?

- [x] Yes
- [ ] No

If any component is using `modalTitle` or `modalSize` as part of their object binding to component, it may cause undesirable results.

## Other information

N/A

Issue #313, opened 2019-11-07
[Chore] Modal Design Adjustments


grahamhency:
### What general chore or task would you like to see performed?
The modal is in need of some design improvements.

###### Example 
<img width="838" alt="Screen Shot 2019-11-07 at 2 12 36 PM" src="https://user-images.githubusercontent.com/11966331/68419808-f2cd7880-0168-11ea-82dd-fea72432cf56.png">

### Are there any specific steps you would like completed?
- [ ] Ensure the title and `x` to close are spaced appropriately per the design
- [ ] Currently, there is a max-width on the modal. It should be increased.

patryk0605:
@grahamhency I'll take the issue

Fagiolia1:
The expanded modal should have a width of 800px.

patryk0605:
Modal specs: https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d6ecbaf4ed8142b75ae33f3

Fagiolia1:
Normal modals should have a min-width of 500px. The XL Modal should have a min-width/max-width of 800px.

patryk0605:
@grahamhency as a FYI, this issue became more of a "Feature" instead of a "Chore" after sitting down with @Fagiolia1 to understand the requirements. I noted the PR as such. The real ask from this issue was to support a title for the modal and also standardize the size of the modal to lg (default) and xl.

grahamhency:
Closed with #333 and #344

PR #333, opened 2019-11-27
Fix modal sizes and add additional documentation for modal sizes

patryk0605:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/mobi/goponents/issues/313

An extension of changes made in PR https://github.com/mobi/goponents/pull/321

## What is the new behavior?

Currently `lg` always expands to the size of `xl` since the `max-width` was set equal to the size of `xl`. This is something I accidentally introduced in https://github.com/mobi/goponents/pull/321 😄 but luckily caught it before it was released. 

Also added additional documentation to showcase the different available sizes.  

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/projects/go-style-guide/src/app/features/ui-kit/components/modal-docs/modal-docs.component.ts
+++ b/projects/go-style-guide/src/app/features/ui-kit/components/modal-docs/modal-docs.component.ts
@@ -13,8 +13,8 @@ export class ModalDocsComponent {
   pageTitle: string = 'Modal';
 
   componentBindings: string = `
-  @Input() modalTitle:  string      = '';
-  @Input() modalSize:   'lg' | 'xl' = 'lg';
+  @Input() modalTitle: string = '';
+  @Input() modalSize: 'lg' | 'xl' = 'lg';
   `;
 
   appModuleImport: string = `
@@ -73,7 +73,7 @@ export class ModalDocsComponent {
   ex_ModalDocsHtml: string = `<go-button (handleClick)="openModal()">Click Me</go-button>`;
 
   ex_ModalDocsOpenLgModal: string = `
-  this.goModalService.openModal(ModalTestComponent, { modalTitle: 'LG Modal (Default)', modalSize: 'lg', content: 'This is a lg modal' });
+  this.goModalService.openModal(ModalTestComponent, { modalTitle: 'LG Modal (Default)', content: 'This is a lg modal' });
   `;
 
   ex_ModalDocsOpenXlModal: string = `
@@ -87,7 +87,7 @@ export class ModalDocsComponent {
   }
 
   openLgModal(): void {
-    this.goModalService.openModal(ModalTestComponent, { modalTitle: 'LG Modal (Default)', modalSize: 'lg', content: 'This is a lg modal' });
+    this.goModalService.openModal(ModalTestComponent, { modalTitle: 'LG Modal (Default)', content: 'This is a lg modal' });
   }
 
   openXlModal(): void {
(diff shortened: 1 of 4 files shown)


projects/go-style-guide/src/app/features/ui-kit/components/modal-docs/modal-docs.component.ts
```diff
@@ -12,6 +12,11 @@ export class ModalDocsComponent {
 
   pageTitle: string = 'Modal';
 
+  componentBindings: string = `
+  @Input() modalTitle:  string      = '';
+  @Input() modalSize:   'lg' | 'xl' = 'lg';
```
StevenUlmer: For consistency sake on the docs should we not line up the '=' here?

projects/go-style-guide/src/app/features/ui-kit/components/modal-docs/modal-docs.component.ts
```diff
@@ -67,10 +72,26 @@ export class ModalDocsComponent {
 
   ex_ModalDocsHtml: string = `<go-button (handleClick)="openModal()">Click Me</go-button>`;
 
+  ex_ModalDocsOpenLgModal: string = `
+  this.goModalService.openModal(ModalTestComponent, { modalTitle: 'LG Modal (Default)', modalSize: 'lg', content: 'This is a lg modal' });
```
StevenUlmer: we don't need to pass in a modalSize here correct because it defaults to `'lg'`?

projects/go-style-guide/src/app/features/ui-kit/components/modal-docs/modal-docs.component.ts
```diff
@@ -67,10 +72,26 @@ export class ModalDocsComponent {
...
+  `;
+
   constructor(private goModalService: GoModalService) { }
 
   openModal(): void {
     this.goModalService.openModal(ModalTestComponent, { modalTitle: 'Example Title', modalSize: 'xl', content: 'This is a modal!' });
   }
 
+  openLgModal(): void {
+    this.goModalService.openModal(ModalTestComponent, { modalTitle: 'LG Modal (Default)', modalSize: 'lg', content: 'This is a lg modal' });
```
StevenUlmer: modalSize should default to `'lg'`, so we shouldn't need to pass it in

projects/go-style-guide/src/app/features/ui-kit/components/modal-docs/modal-docs.component.ts
patryk0605: I noticed on some doc pages we line up the `=` and on some we don't. If we've agreed on not lining them up, then I will change it.

projects/go-style-guide/src/app/features/ui-kit/components/modal-docs/modal-docs.component.ts
patryk0605: correct that it's not mandatory to pass, but I'm passing it here just in case we ever change the default. Less to worry about updating later, however if you feel it would make more sense to remove the `modalSize` here then I can.

patryk0605: correct that it's not mandatory to pass, but I'm passing it here just in case we ever change the default. Less to worry about updating later, however if you feel it would make more sense to remove the `modalSize` here then I can.

patryk0605:
@StevenUlmer I've made the requested changes - PTAL

Issue #305, opened 2019-11-06
[Feature] Header Bar for "Interior Pages"


Fagiolia1:
### What is the feature that you would like to propose?
We need a top header bar with an H2 title on the left and the option for buttons on the right. 

This header bar should be fixed to the top of the page and follow the user as they scrolls down the page.

Here is the DSM information regarding the header bar: 

https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d7a956118677478a3a5f60c

We would also like the Header Bar to have an option for a back button. This back button would function as a browser back for interior pages within an application. 

For example, if a user selects an invoice from a table of invoices, the Header Bar on the individual invoice page would have a back button attached. This would allow the user to navigate quickly back to the list of invoices instead of having to use the left navigation or their native browser back.

Example of back button on the Header Bar:

https://tangoe.invisionapp.com/share/YJTHRTCFA4X#/382017621_Invoice_Task

grahamhency:
@Fagiolia1 thanks for submitting this.

Are we dropping the "back" functionality requirement?

Fagiolia1:
@grahamhency We'd like to have the back button there as an option. Is it possible this can be configurable? Some screens right now are still going to require this.

grahamhency:
@Fagiolia1 I think so, would you mind adding the specifics of how that thing should work in the description of the task please?

jaykravetz:
I can work on this one

jaykravetz:
@Fagiolia1 Should the back arrow use the browser history or should there be an option to set the go-back-to path (or both)?

Fagiolia1:
@jaykravetz at the moment lets keep browser history. We're considering other ways to display this pages in the future, but for now we're sticking with browser history.

grahamhency:
@Fagiolia1 how about getting rid of it??? :smirk:

What happens if there is no history for the tab?

eligolding:
>how about getting rid of it???

Best. Idea. Ever. 

>What happens if there is no history for the tab?

What I've been doing in my local version of this component in Command Center is pass in a fallback url.

Fagiolia1:
> @Fagiolia1 how about getting rid of it??? 😏
> 
> What happens if there is no history for the tab?

If there is no history, the back arrow should drop.

eligolding:
>the back arrow should drop

[...]

PR #336, opened 2019-12-02
Page header bar

jaykravetz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Closes: #305 

## What is the new behavior?
Adds a Page Header Bar component with a back button and action buttons

- [ ] Yes
- [x] No


projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
+@import "../../../styles/_variables.scss";
+.go-header-bar {
+  margin: -2rem;
```
grahamhency: We can use the `$column-gutter--double` variable for this, you'll still need to make it negative though.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  height: 68px;
```
grahamhency: We definitely do not want to assign a fixed height here. This header should be flexible based on the content within it.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  background: white;
```
grahamhency: We can use the `$base-light` variable for this.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  box-shadow: 0 0 2px rgba(0, 0, 0, 0.12), 0 2px 2px rgba(0, 0, 0, 0.24);
```
grahamhency: We should try to utilize one of the `box-shadow` variables for this. Check the available variables in the main `variables.scss` file to see if any of them would work. If none of them match, check with UX to see which shadow we should use.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
+  position: fixed;
+}
+
+.go-page-header__title {
+  padding: 0 2rem;
```
grahamhency: We can use the `$column-gutter--double` variable for the `2rem` here.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  margin-bottom: -0.25rem;
```
grahamhency: We can use the `$column-gutter--quarter` variable here. You'll still need to make it negative.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+}
+
+.go-page-header__back-arrow {
+  cursor: pointer;
+  align-items: center;
+  padding: 1.5rem 1rem;
```
grahamhency: We should make use of the `$column-gutter` variables for these units.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  margin-right: -1rem;
```
grahamhency: We should make use of the `$column-gutter` variable here.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  border-right: solid 1px lighten($base-light-secondary, 17%);
```
grahamhency: Do we need to use this `lighten()` here? Could we use the `$theme-light-border` variable for this?

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  flex-basis: 40px;
```
grahamhency: Could we utilize one of the `$column-gutter` variables for this? Or do we really need this item?

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  color: $base-dark;
+  font-size: 18px;
```
grahamhency: @jaykravetz we make use of `rems` for all units of measure in Goponents. We'll need to switch this to `rems`.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,39 @@
...
+.go-header-bar__back-arrow--disabled {
+  color: $base-light-secondary;
+  cursor: default;
+}
+
+.go-page-header__actions {
+  display: flex;
+  justify-content: flex-end;
+  flex: 1;
+  padding-right: 2rem;
```
grahamhency: We can switch this to the `$column-gutter--double` variable.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  backButtonDisabled(): boolean {
+    return !this.goBackFn && window.history.length === 1;
+  }
+
+  getHeight(): string {
+    return this.headerBar.nativeElement.offsetHeight.toString() + 'px';
+  }
+
+  getWidth(): string {
+    return this.headerBar.nativeElement.parentElement.parentElement.offsetWidth.toString() + 'px';
```
grahamhency: While this does work, it assumes a specific structure in the HTML. This will likely be prone to bugs if the implementor does not put the `<go-header-bar>` at the root level inside of the routes component. 

This works:
```
<go-header-bar>
</go-header-bar>

<div class="go-container">
  page content
</div>
```

This will break:
```
<div>
  <go-header-bar>
  </go-header-bar>
</div>

<div class="go-container">
  page content
</div>
```

projects/go-tester/src/app/app.module.ts
```diff
@@ -93,7 +94,8 @@ import { AppGuard } from './app.guard';
     GoTableModule,
     GoTextAreaModule,
     GoToastModule,
-    GoToasterModule
+    GoToasterModule,
+    GoHeaderBarModule
```
grahamhency: Can this be alphabetized, please!

projects/go-tester/src/app/components/test-page-1/test-page-1.component.html
```diff
@@ -1,3 +1,8 @@
+<go-header-bar title="This is the title" [showBackArrow]="true">
+  <go-button go-header-bar-action>Hello</go-button>
```
grahamhency: Does each one of these `go-header-bar-action` items get rendered into the `ng-content` of the component? Also, it would probably be easier to format these using the button group classes, https://tangoe.design/ui-kit/button#button-group

I know this is just the test project, but still.

projects/go-tester/src/app/app.module.ts
eligolding: In case you need help: https://www.youtube.com/watch?v=qTvhKZHAP8U

jaykravetz: I watched it 6 times but the whole ACDBs still confuse me

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.ts
jaykravetz: I injected the `ElementRef` but there is still a styling issue when wrapping in a div:

```html
<go-header-bar>
</go-header-bar>
```

Results in correct styling:
image


```html
<div>
    <go-header-bar>
    </go-header-bar>
</div>
```

results in incorrect style:
image

grahamhency: Hmm, we might have to go with a slightly different approach. We could query for the closest element with the `go-layout__route-container-outlet` class. This would look like:

```
getWidth(): string {
  return this.elemRef.nativeElement.closest('.go-layout__route-container-outlet').offsetWidth.toString() + 'px';
}
```

grahamhency: Additionally, we'll need may need to recalculate this width based on window resize. Angular might already know to call these methods based so you'll want to double check that.

eligolding: Another approach I was thinking of was to use a [portal](https://material.angular.io/cdk/portal/overview) to render this into the right slot in the layout. That way regardless of where in the template you put this, it will render in a place that we already know the dimensions and we don't have to do all these calculations.

jaykravetz: `this.elemRef.nativeElement.closest('.go-layout__route-container-outlet');` returns `null`. would it be a problem to query for it locally?

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        showBackArrow: true
+      };
+    });
+
+    it('calls history.back when no observers exist', () => {
+      spyOn(window.history, 'back');
+
+      component.processBackBtn();
+
+      expect(window.history.back).toHaveBeenCalledTimes(1);
```
grahamhency: This could probably be simplified to `.toHaveBeenCalled()`

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+    it('calls custom method when observer is passed', () => {
+      const handlerSpy: any = createSpy('handler', (): void => {});
+
+      component.currentItem.backButtonFn.subscribe(handlerSpy);
+      component.processBackBtn();
+
+      expect(handlerSpy).toHaveBeenCalledTimes(1);
```
grahamhency: This could probably be simplified to `.toHaveBeenCalled()`

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.scss
```diff
@@ -0,0 +1,40 @@
+@import "../../../styles/variables";
+
+.go-header-bar {
```
grahamhency: We'll also need to add `positive: relative` to this to properly position it

Issue #363, opened 2019-12-11
[BUG] incorrect datepicker font-color for dark mode


jaredami:
**Describe the bug**
Some of the font-colors for our `go-datepicker` inherit from their parent and thus are turned white when their parent uses our dark theme, making the font illegible on the white background of the datepicker.

**To Reproduce**
1. Add a `go-datepicker` within an element that uses our dark theme (e.g. a `go-off-canvas`).
2. Open the calendar view of the datepicker to see the problem.

**Expected behavior**
The font-colors of the datepicker should be set to black so that they are always visible on the white background of the datepicker.

**Screenshots**
Current behavior:
![Screen Shot 2019-12-11 at 1 35 26 PM](https://user-images.githubusercontent.com/23343417/70650049-56413f00-1c1c-11ea-8470-c5056e341d1d.png)

Desired behavior:
![Screen Shot 2019-12-11 at 1 36 50 PM](https://user-images.githubusercontent.com/23343417/70650019-49245000-1c1c-11ea-8a62-6eb51b00536d.png)

grahamhency:
Closed with #364

PR #364, opened 2019-12-11
Fix datepicker font-colors for dark theme

jaredami:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Some of the font-colors for our go-datepicker inherit from their parent and thus are turned white when their parent uses our dark theme, making the font illegible on the white background of the datepicker.

Screen Shot 2019-12-11 at 1 35 26 PM

Issue Number: #363 

## What is the new behavior?
The font-colors of the datepicker are now set to black so that they are always visible on the white background of the datepicker.

Screen Shot 2019-12-11 at 1 36 50 PM

- [ ] Yes
- [ x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-datepicker/calendar-views.scss
+++ b/projects/go-lib/src/lib/components/go-datepicker/calendar-views.scss
@@ -1,6 +1,7 @@
 @import '../../../styles/variables';
 
 .go-calendar__header {
+  color: $theme-light-color;
   display: inline-flex;
   height: 1rem;
   margin-top: 0.5rem;
@@ -69,5 +70,6 @@
 }
 
 .go-calendar__table {
+  color: $theme-light-color;
   width: 100%;
 }


projects/go-lib/src/lib/components/go-datepicker/calendar-views.scss
```diff
@@ -1,6 +1,7 @@
 @import '../../../styles/variables';
 
 .go-calendar__header {
+  color: $base-dark;
```
StevenUlmer: These base colors are used to set up our theme colors. Should we use our theme color here instead? `$theme-light-color`

jaredami: @StevenUlmer Gotcha. So do we *always* want to reference the theme colors rather than the base colors?

StevenUlmer: Almost always the answer to that is yes. There might be rare cases where we are doing something outside of the scope of a theme, but here this text should be the same as whatever the light theme is and if that themes color ever changes, then we want this to change as well.

Issue #234, opened 2019-10-03
[Chore] Update Radio Button Design


grahamhency:
### What general chore or task would you like to see performed?
The design of our radio buttons need to be updated to meet the design spec. 

###### Example 
https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5d3b2fe780452ee15cc785fb

### Are there any specific steps you would like completed?
1. Each version of the radio button should be represented in our component

grahamhency:
@patryk0605 is working on this.

grahamhency:
Closed with #399

PR #399, opened 2020-01-07
Update go-radio styles + address bug with multiple radio groupings

patryk0605:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Style Changes

Issue Number: https://github.com/mobi/goponents/issues/234

This PR addresses two things:

1. Update radio button styles per DSM guidelines
2. Fix a bug where if there were multiple radio groups with the same value, the value would be selected in all groups. This issue can be seen today in the design system example. Issue was occurring because a unique name per group was missing. I'm creating a name using a random generator instead of asking for a key to avoid having any new inputs to the component as well as support backward compatibility.

## What is the new behavior?

1. Radio button styles updated
2. `name` added to each radio grouping

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
+++ b/projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
@@ -1,0 +1,38 @@
+@import '../../../styles/variables';
+
+input[type='radio'] {
+  display: none;
+}
+
+.go-radio__custom-radio {
+  display: inline-block;
+  width: 13px;
+  height: 13px;
+  border-radius: 100%;
+  cursor: pointer;
+  margin: 3px 3px 0px 5px;
+  border: 1px solid lighten($base-light-secondary, 13.5%);
+  background-color: $base-light;
+}
+
+input[type='radio']:enabled:hover + .go-radio__custom-radio {
+  box-shadow: 0 0 2px 1.5px lighten($base-light-secondary, 13.5%);
+}
+
+input[type='radio']:enabled:checked + .go-radio__custom-radio {
+  border: 4px solid $ui-color-primary;
+  background-color: $base-light;
+}
+
+input[type='radio']:enabled:checked:hover + .go-radio__custom-radio {
+  box-shadow: 0 0 2px 1.5px $ui-color-primary-hover;
+}
+
+input[type='radio']:disabled + .go-radio__custom-radio {
+  cursor: not-allowed;
+  background-color: $theme-light-app-bg;
+}
+
+input[type='radio']:disabled:checked + .go-radio__custom-radio {
+  border: 4px solid lighten($base-light-secondary, 13.5%);
+}
--- a/projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
+++ b/projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
@@ -1,13 +1,13 @@
 @import '../../../styles/variables';
 
-input[type='radio'] {
+.go-radio__input {
   display: none;
 }
 
 .go-radio__custom-radio {
   display: inline-block;
-  width: 13px;
-  height: 13px;
+  width: .75rem;
+  height: .75rem;
   border-radius: 100%;
   cursor: pointer;
   margin: 3px 3px 0px 5px;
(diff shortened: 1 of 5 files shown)


projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
```diff
@@ -0,0 +1,38 @@
+@import '../../../styles/variables';
+
+input[type='radio'] {
```
grahamhency: Instead of using the element selector in these, a BEM class should be used. Something like `.go-radio__input` would suffice.

projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
```diff
@@ -0,0 +1,38 @@
...
+  display: none;
+}
+
+.go-radio__custom-radio {
+  display: inline-block;
+  width: 13px;
+  height: 13px;
```
grahamhency: We use `rems` all throughout, would you mind converting these to `.75rem`?

projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
```diff
@@ -0,0 +1,38 @@
...
+  border-radius: 100%;
+  cursor: pointer;
+  margin: 3px 3px 0px 5px;
+  border: 1px solid lighten($base-light-secondary, 13.5%);
+  background-color: $base-light;
+}
+
+input[type='radio']:enabled:hover + .go-radio__custom-radio {
+  box-shadow: 0 0 2px 1.5px lighten($base-light-secondary, 13.5%);
```
grahamhency: This seems like an aggressive box-shadow. Also, we're introducing yet another box shadow. Can we make use of one of the ones that already exist in our `variables` file?

projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
```diff
@@ -0,0 +1,38 @@
...
+}
+
+input[type='radio']:enabled:checked + .go-radio__custom-radio {
+  border: 4px solid $ui-color-primary;
+  background-color: $base-light;
+}
+
+input[type='radio']:enabled:checked:hover + .go-radio__custom-radio {
+  box-shadow: 0 0 2px 1.5px $ui-color-primary-hover;
```
grahamhency: Same question here on the box-shadow

projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
patryk0605: This is the design Kim signed off on, so yes we are introducing another size 😄. I experimented with the ones we have predefined and they don't match the DSM specs or come close to it. Also the predefined box-shadows we have don't play nicely with a circular object like a radio button.

projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
patryk0605: updated

projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
patryk0605: updated

projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
patryk0605: same answer as above

patryk0605:
@grahamhency I made the selector and rem conversion changes you requested. For the box-shadows, those are the ones Kim approved for the design so not sure what you want to do about them.

projects/go-lib/src/lib/components/go-radio/go-radio-button.component.scss
grahamhency: Okay. Let's get her to take a look at these in the preview for sign off. I'll send it her way.

kpmurphh: This looks good to me. The box shadow is looking a little excessive right now, but I have submitted issues to replace both the button box shadow and the form controls (which includes radios, checkboxes and switches) box shadow from this to what is documented in the DSM and the issues. I imagine that this can be approved without changing the box shadow since that's a CSS issue and not hardcoded to the radio?

grahamhency: Thanks Kim! We can handle changing this in the other PR's for form controls and buttons.

Issue #402, opened 2020-01-10
[BUG] Change detection in GoDatePicker seems to not be working


jaykravetz:
**Describe the bug**
Calling `setValue` on a FormControl that is bound to a `go-datepicker` does not update the datepicker's value 

**To Reproduce**
1. Add go-datepicker to any component
2. Inside an async callback (`setTimeout`, etc), call `setValue` on the datepicker's `FormControl`
3. The datepicker field will not be updated

**Expected behavior**
Calling `setValue` on a datepicker's `FormComponent` should update the datepickers value and display the date in the view

**Screenshots**
See example [test case](https://github.com/jaykravetz/goponents/commit/2ea45726d0a4639e5b6fffc9e763b00f326cf024#diff-4feeb60806a53a6ff9ee7a9a6bb20e89R1)

patryk0605:
I'll take this

grahamhency:
Closed with #405

PR #405, opened 2020-01-13
[Bug] Fix change detection bug in date picker 

patryk0605:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Date picker control was being update but display value was not.

Issue Number: https://github.com/mobi/goponents/issues/402

## What is the new behavior?

Display value is updated per value change.

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/projects/go-lib/src/lib/components/go-datepicker/go-datepicker.component.ts
+++ b/projects/go-lib/src/lib/components/go-datepicker/go-datepicker.component.ts
@@ -54,7 +54,7 @@ export class GoDatepickerComponent implements OnDestroy, OnInit {
     this.subscription = this.control.valueChanges.subscribe((value: Date) => {
       if (!value && this.selectedDate) {
         this.control.setErrors([{ message: 'format is invalid' }]);
-      } else {
+      } else if (value) {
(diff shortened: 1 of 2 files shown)


projects/go-lib/src/lib/components/go-datepicker/go-datepicker.component.ts
```diff
@@ -54,6 +54,8 @@ export class GoDatepickerComponent implements OnDestroy, OnInit {
     this.subscription = this.control.valueChanges.subscribe((value: Date) => {
       if (!value && this.selectedDate) {
         this.control.setErrors([{ message: 'format is invalid' }]);
+      } else {
```
StevenUlmer: I think we want `else if (value)` here. The reason is the control could be set to null by the datePicked method, because it's out of range, and we don't want to clear the field if they enter an invalid date, just show them it's not valid.

patryk0605: Ah yes, good point. I've made the change.

patryk0605:
@StevenUlmer I've made the requested change, PTAL

PR #460, opened 2020-02-06
[Bug] Accordion Panel Content Positioning

grahamhency:
## PR Checklist

## PR Type
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The `accordion-panel` has the ability to project content into the header. Currently, the content was being forced to the right every time.

## What is the new behavior?
The `accordion-panel` now projects content into the available space in the header

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.html
+++ b/projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.html
@@ -21,9 +21,9 @@
       <span class="go-accordion-panel__title-text">{{ heading }}</span>
     </div>
     
-    <ng-container *ngIf="headerContent">
+    <div class="go-accordion-panel__projection">
       <ng-container *ngTemplateOutlet="headerContent"></ng-container>
-    </ng-container>
+    </div>
 
     <go-icon
       class="go-accordion-panel__control"
--- a/projects/go-style-guide/src/app/features/ui-kit/components/accordion-docs/components/accordion-panel-docs/accordion-panel-docs.component.html
+++ b/projects/go-style-guide/src/app/features/ui-kit/components/accordion-docs/components/accordion-panel-docs/accordion-panel-docs.component.html
@@ -214,11 +214,7 @@
           </go-accordion-panel>
           <go-accordion-panel heading="Projected Heading">
             <ng-template #headerContent>
-              <div class="go-container go-container--justify-end">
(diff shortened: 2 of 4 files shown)


projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.html
```diff
@@ -21,9 +21,9 @@
       <span class="go-accordion-panel__title-text">{{ heading }}</span>
     </div>
     
-    <ng-container *ngIf="headerContent">
+    <div class="go-accordion-panel__projection">
```
jaredami: Would it be best to avoid rendering an empty div when there is no `headerContent`? Or does that not matter?

projects/go-style-guide/src/app/features/ui-kit/components/accordion-docs/components/accordion-panel-docs/accordion-panel-docs.component.html
```diff
@@ -214,11 +214,7 @@ <h4 class="go-heading-4 go-heading--underlined">View</h4>
           </go-accordion-panel>
           <go-accordion-panel heading="Projected Heading">
             <ng-template #headerContent>
-              <div class="go-container go-container--justify-end">
-                <div class="go-column go-column--no-padding">
-                  <go-button>Projected</go-button>
-                </div>
-              </div>
+              <go-pill [removable]="false">Projected</go-pill>
```
jaredami: Should we update the `headerTemplateExample` code snippet referenced below with this change?

projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.html
grahamhency: We actually want to do this. The `div` will collapse if there is no content inside to fit the content from the other flex items, but it will grow to fill the available space as much as it can.

projects/go-style-guide/src/app/features/ui-kit/components/accordion-docs/components/accordion-panel-docs/accordion-panel-docs.component.html
grahamhency: Yes we should, good catch. I updated that.

PR #485, opened 2020-02-17
[Bug] Table Search Subscription

grahamhency:
## PR Checklist

## PR Type
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The `searchTerm` on the table search is being subscribed to on every `onChange` event. Causing an extreme amount of excess events to be called the more the table search is used.

## What is the new behavior?
Moved the `searchSetup` method into the `ngOnInit` instead of in `ngOnInit` so that we only subscribe to the changes once.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.ts
@@ -81,7 +81,10 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
     if (!this.tableConfig) {
       throw new Error('GoTableComponent: tableConfig is a required Input');
     } else {
-      this.renderTable();
+      // we have to do call setupSearch here because it creates a subscription
+      // if we call it in ngOnChanges it will create a new subscription
+      // everytime ngOnChanges is triggered, which is not good
+      this.setupSearch();
     }
   }
 
@@ -103,7 +106,6 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
       this.allData = this.localTableConfig.tableData;
       this.setTotalCount();
       this.handleSort();
-      this.setupSearch();
       this.setPage(this.localTableConfig.pageConfig.offset);
     }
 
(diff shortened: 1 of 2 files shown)


projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -353,7 +355,8 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
     return this.localTableConfig.dataMode === GoTableDataSource.server;
   }
 
-  private tableChangeOutcome(): void {
+  private tableChangeOutcome(event: string): void {
+    // console.log(event);
```
StevenUlmer: should we remove this?

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -416,8 +419,9 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
       this.localTableConfig.searchConfig.searchTerm = searchTerm;
       if (!this.isServerMode()) {
         this.performSearch(searchTerm.toLowerCase());
+      } else {
```
StevenUlmer: Should we emit this regardless of whether it's serverMode or not. Everywhere else regardless of whether it's the server mode or not I believe we're still emitting that something changed.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -353,7 +355,8 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
     return this.localTableConfig.dataMode === GoTableDataSource.server;
   }
 
-  private tableChangeOutcome(): void {
+  private tableChangeOutcome(event: string): void {
```
StevenUlmer: should we emit this event with the tableChange, or do we need to have a param here?

grahamhency: We actually don't need this param. This was from testing. Removed.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
grahamhency: Yep, lol. Removed.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
grahamhency: We actually should not. It will duplicate the event because we call `tableChangeOutcome` within `performSearch`.

PR #538, opened 2020-04-20
[Bug] Copy Button Overflow

grahamhency:
## PR Checklist

## PR Type
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Inside of a table cell, the current overflow implementation does not work.

## What is the new behavior?
Implemented absolute positioning on the host element to ensure proper overflow.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-copy/go-copy.component.scss
+++ b/projects/go-lib/src/lib/components/go-copy/go-copy.component.scss
@@ -1,10 +1,14 @@
-.go-copy__textarea {
-  position: absolute;
-  left: 99999px;
+:host.go-copy {
+  position: relative;
 }
 
-.go-copy {
-  overflow: hidden;
-  position: relative;
+.go-copy__textarea {
+  left: 99999px;
+  position: absolute;
+}
+
+.go-copy__content {
   cursor: pointer;
+  overflow: hidden;
+  position: absolute;
 }
--- a/projects/go-lib/src/lib/components/go-copy/go-copy.component.ts
+++ b/projects/go-lib/src/lib/components/go-copy/go-copy.component.ts
@@ -1,4 +1,4 @@
-import { Component, ElementRef, Input, ViewChild } from '@angular/core';
+import { Component, ElementRef, HostBinding, Input, ViewChild } from '@angular/core';
 
 @Component({
   animations: [],
@@ -7,6 +7,9 @@ import { Component, ElementRef, Input, ViewChild } from '@angular/core';
   styleUrls: ['./go-copy.component.scss']
 })
 export class GoCopyComponent {
+
+  @HostBinding('class.go-copy')
+
   @Input() text: string;
 
   @ViewChild('copyText') copyText: ElementRef;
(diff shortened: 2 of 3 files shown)


projects/go-lib/src/lib/components/go-copy/go-copy.component.scss
```diff
@@ -1,10 +1,14 @@
...
 .go-copy__textarea {
-  position: absolute;
   left: 99999px;
+  position: absolute;
 }
 
-.go-copy {
-  overflow: hidden;
-  position: relative;
+.go-copy__content {
```
StevenUlmer, line 10: Is there an html element that has this class on it?

projects/go-lib/src/lib/components/go-copy/go-copy.component.ts
```diff
@@ -7,6 +7,9 @@ import { Component, ElementRef, Input, ViewChild } from '@angular/core';
   styleUrls: ['./go-copy.component.scss']
 })
 export class GoCopyComponent {
+
+  @HostBinding('class') class: string = 'go-copy';
```
StevenUlmer, line 11: Will this prevent someone from adding their own class to the component? Should we set it like this instead?
```
@HostBinding('class.go-copy')
```

projects/go-lib/src/lib/components/go-copy/go-copy.component.scss
```diff
@@ -1,10 +1,14 @@
+.go-copy {
```
StevenUlmer, line 1: since this is being added to the host do we have to specify this as a host attribute?
```
:host.go-copy {
  position: relative;
}
```

projects/go-lib/src/lib/components/go-copy/go-copy.component.scss
grahamhency: Yeah, I forgot to update that so good catch. I added this class to the appropriate element in the template file.

projects/go-lib/src/lib/components/go-copy/go-copy.component.scss
grahamhency: We don't! This is what the `@HostBinding()` allows us to do. We can keep the scoped styles in the template without having to use `:host`.

projects/go-lib/src/lib/components/go-copy/go-copy.component.ts
grahamhency: Yeah, this is a good catch as well. I switched this around. Thanks

projects/go-lib/src/lib/components/go-copy/go-copy.component.scss
StevenUlmer: @grahamhency looks like while go-copy is currently being added to the host property, the css applied here is not being applied to the host property in the deploy preview.

grahamhency: Well shit, you're right. I remember not needing to add that at some point... perhaps I remembered wrong. :shrug: I added the `:host` portion in the SCSS.

PR #543, opened 2020-04-30
Removes text decoration from go-link

StevenUlmer:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Interface Style Changes

Currently using the go-link class from goponents does not produce a link that looks the same as links in the style guide do. This is because in the style guide we have a style in there that removes the text decoration from all link tags.

## What is the new behavior?
We remove the bottom border from go-link and add text decoration to make all the same styles are being applied to link tags in the style guide.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/styles/_typography.scss
+++ b/projects/go-lib/src/styles/_typography.scss
@@ -29,9 +29,9 @@ body {
 }
 
 .go-link {
-  border-bottom: 1px solid;
   color: $theme-light-color;
   display: inline-block;
+  text-decoration: underline;
   transition: $global-transition-duration $global-transition-timing;
 
   &:active,


StevenUlmer:
> Thanks for doing this, but I forgot about something regarding this task that I realized a while back. We actually should switch this around to remove the `border-bottom` and use `text-decoration: underline` instead.
> 
> The `border-bottom` can cause some weird spacing issues if there is text or content immediately below the link.

Ya that makes sense, I've switched it to remove the `border-bottom` and add `text-decoration`

PR #546, opened 2020-05-05
[Chore] Search Updates

grahamhency:
## PR Checklist

## PR Type
- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The search bar in the middle section of the header that threw an error in a compiled application.

## What is the new behavior?
Fixed this issue, as well as implemented a few other adjustments and updates to the search.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-header/go-header.component.ts
+++ b/projects/go-lib/src/lib/components/go-header/go-header.component.ts
@@ -17,8 +17,6 @@ export class GoHeaderComponent implements OnChanges, OnDestroy {
   @Input() logoLink: string = '';
   @Input() logoLinkUseRouterLink: boolean = true;
 
-  @ViewChild('middleSection') middleSection: ElementRef;
-
   public brandColor: string;
   public brandColorIsDark: boolean;
 
@@ -70,10 +68,6 @@ export class GoHeaderComponent implements OnChanges, OnDestroy {
 
   toggleSideMenu(): void {
     this.sideNavService.toggleNav();
-  }
-
-  middleContentExists(): boolean {
-    return this.middleSection && this.middleSection.nativeElement.childElementCount > 0;
   }
 
   private setupResizeSubscription(): void {
--- a/projects/go-lib/src/lib/components/go-search/go-search.component.ts
+++ b/projects/go-lib/src/lib/components/go-search/go-search.component.ts
@@ -56,7 +56,6 @@ export class GoSearchComponent implements OnInit {
         this.goSearchService.showNoResultsMessage = false;
         this.goSearchService.hasResults = false;
         this.goSearchService.showLoader = false;
-        this.goSearchService.updateSearchTerm();
       }
     });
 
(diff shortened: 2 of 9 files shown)


projects/go-lib/src/lib/components/go-header/go-header.component.ts
```diff
@@ -72,10 +70,6 @@ export class GoHeaderComponent implements OnChanges, OnDestroy {
     this.sideNavService.toggleNav();
   }
 
-  middleContentExists(): boolean {
-    return this.middleSection && this.middleSection.nativeElement.childElementCount > 0;
-  }
```
grahamhency, line 77: This is totally unnecessary and was causing issues, thus I am removing it.

projects/go-lib/src/lib/components/go-search/go-search.component.ts
```diff
@@ -75,6 +80,18 @@ export class GoSearchComponent implements OnInit {
...
+    if (event.code === 'ArrowDown') {
+      this.goSearchService.arrowDownEvent.emit();
+    }
+  }
+
+  focusOnSearch(): void {
+    this.goSearchService.focusOnSearchEvent.subscribe(() => {
+      this.searchInput.nativeElement.focus();
+    });
+  }
```
StevenUlmer, line 93: What was the purpose behind adding this here? Was it just to apply the correct css?

projects/go-lib/src/lib/components/go-search/go-search.component.ts
```diff
@@ -75,6 +80,18 @@ export class GoSearchComponent implements OnInit {
     }
   }
 
+  keyDown(event: Partial<KeyboardEvent>): void {
+    if (event.code === 'ArrowDown') {
+      this.goSearchService.arrowDownEvent.emit();
+    }
+  }
```
StevenUlmer, line 87: Should we allow for more keyboard navigation than just arrow key down here? If we're looking for navigation within search results the user should have the ability to go down or up in the results right? Should we also add the ability to close the search with the escape key? Not sure if that works currently, but I would assume it doesn't.

grahamhency:
I believe there is a misunderstanding on how this component is implemented. This component does not handle rendering the search results because formatting could differ vastly between products and implementations.

That being said, the reasoning for adding these keyboard and focus events are so that you can handle the navigation from your results component back into the search component.

`arrowDownEvent` is so that you know when the user navigates down, out of the search input.

`focusOnSearchInput` so for then you need to apply focus back on the search input.

The actual navigation through the results will have to be handled upon on the implementation of this component. Does all of this make sense?

projects/go-lib/src/lib/components/go-search/go-search.component.ts
grahamhency: See my comment below.

projects/go-lib/src/lib/components/go-search/go-search.component.ts
grahamhency: See my comment below, but we could look at doing the escape key as well.

Issue #524, opened 2020-03-24
[BUG] Header Bar Buttons Jump when navigating across pages with Header Bars


Fagiolia1:
**Describe the bug**
When navigating between pages with header bar buttons, the buttons on the header bar jump from left to right once the page fully loads.

**To Reproduce**
1. Go to a page with a header bar and some buttons.
2. Navigate to another page with header and buttons.
3. Watch the buttons move from the right to fully right aligned once the page loads.

**Expected behavior**
The buttons should not be jumping between page loads

**Screenshots**
![image](https://user-images.githubusercontent.com/8883257/77449791-cbe6aa80-6dc8-11ea-8b49-209efa908b50.png)

grahamhency:
@mobi/global-goponents-devs, @mobi/tangoe-ux-ui I cannot seem to replicate this issue in Goponents and the `mobi` app does not yet use this. Can any of you provide an example of where this is happening?

kpmurphh:
@grahamhency I can't replicate it either. I tried a few different ways in the cloud app (where Andy took these screenshots) and everything looked fine. The cloud team also hasn't reported it as a bug so I'm hoping it was fixed indirectly. If it does come up I will definitely let you know.

grahamhency:
Hmm. I think @eligolding was able to recreate it. He reached out to me, any chance you could provide some insight here @eligolding?

Fagiolia1:
@grahamhency @kpmurphh Jason was able to recreate it a few times. You should be able to recreate it within Cloud by navigating from a page that has a header bar to another page with a header bar. 

See if Jason has an example as well. He saw something similar in Telecom.

kpmurphh:
Hey @cosmicdust73 do you know where we can replicate this?

eligolding:
I reproduced it by clicking on a link to an account from the invoice page in Cloud.

grahamhency:
I will look into this and see if maybe I can recreate in the style guide.

grahamhency:
@eligolding, @mobi/tangoe-ux-ui I found the culprit for this and I've got a fix up in a PR. Turns out it revolved around a routing animation being applied to the header bar.

eligolding:
![ ](https://media.giphy.com/media/xT5LMFEB6Z6DADQBP2/giphy.gif)

grahamhency:
Closed with #585

PR #585, opened 2020-06-11
[Bug] Header Bar Actions Shifting

grahamhency:
## PR Checklist

## PR Type
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When switching between routes that have different header bar's, the previous header bar's content would still persist and causing a "jumping" of the new header bar's content.

Issue Number: #524 

## What is the new behavior?
This no longer happens.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.html
+++ b/projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.html
@@ -1,12 +1,22 @@
-<div class="go-header-bar" *ngIf="currentItem">
-  <a *ngIf="currentItem.showBackArrow"
-     (click)="processBackBtn()" class="go-header-bar__back-arrow">
-    <go-icon icon="arrow_back"></go-icon>
-  </a>
-  <h2 class="go-heading-2 go-header-bar__title">{{ currentItem.title }}</h2>
+<div
+  class="go-header-bar"
+  *ngIf="currentItem">
+
+  <go-icon
+    *ngIf="currentItem.showBackArrow"
+    (click)="processBackBtn()"
+    class="go-header-bar__back-arrow"
+    icon="arrow_back">
+  </go-icon>
+
+  <h2
+    class="go-heading-2 go-header-bar__title"
+    [ngClass]="{ 'go-header-bar__title--no-arrow': !currentItem.showBackArrow }">
+    {{ currentItem.title }}
+  </h2>
 
   <div class="go-header-bar__actions">
-    <ng-content select="[go-header-bar-action]"></ng-content>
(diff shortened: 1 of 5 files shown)


projects/go-lib/src/lib/components/go-layout/go-layout.component.html
```diff
@@ -6,13 +6,13 @@
     </nav>
     <div class="go-layout__view-container">
 
-      <go-header-bar #headerBar [@routerAnimation]="routeAnimation(mainRouter)"></go-header-bar>
```
grahamhency, line 9: **This** is actually the culprit of why this issue was happening. Angular was performing route animations on the header bar, but the positioning was off with the header bar. Rather than investigating a clean solution to absolutely positioning the elements inside of the header bar, I opted to just remove the animation for now.

projects/go-lib/src/lib/components/go-layout/go-layout.component.scss
```diff
@@ -1,4 +1,5 @@
 @import '../../../styles/variables';
+@import '../../../styles/mixins';
```
grahamhency, line 2: This addition is actually needed, it fixes an issue where the `z-index` on the loader never get's set to the proper value because the mixin for `z-index()` was never imported.

projects/go-lib/src/lib/components/go-header-bar/go-header-bar.component.html
```diff
@@ -1,12 +1,22 @@
...
+  </go-icon>
+
+  <h2
+    class="go-heading-2 go-header-bar__title"
+    [ngClass]="{ 'go-header-bar__title--no-arrow': !currentItem.showBackArrow }">
+    {{ currentItem.title }}
+  </h2>
 
   <div class="go-header-bar__actions">
-    <ng-content select="[go-header-bar-action]"></ng-content>
```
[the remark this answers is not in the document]
StevenUlmer: Ya I agree that this is not needed. We might want to add a deprecation warning on this PR just in case someone was using it, but it's not in the docs, so it shouldn't be used anywhere.

grahamhency:
> This looks good to me. The only thing I had was potentially adding a deprecation warning on the next release just in case someone was using the `go-header-bar-action` in conjunction with the header bar, but everything else looks good

Good thinking, but it could not have been used anywhere that would not have been broken. 

Because it looks for `go-header-bar-action` inside of where the `go-header-bar` would be rendered and the only place that the `go-header-bar` is rendered is inside of the `go-layout` (which is not editable outside of Goponents) then we can assume it's not used anywhere else.

Even if someone did implement the `go-header-bar` outside of the `go-layout`, that probably would've been broken regardless. The positioning, and directive, would've broken the layout.

PR #568, opened 2020-06-02
[Bug] Fix Icon Only go-buttons

grahamhency:
## PR Checklist
Please check if your PR fulfills the following requirements:
- [x] The commit message follows our guidelines: https://github.com/mobi/goponents/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?
Issue Number: #563 

## What is the new behavior?
If only the `buttonIcon` is set and no content exists inside of a `go-button` then appropriate padding and alignment is applied to the button for just an icon.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

Issue #563, opened 2020-06-01
[BUG] goButton with Only Icon (No Text) has extra margin next to icon


Ashygoyal:
**Describe the bug**
In order to have a button with only icon (and no text), if we use `<go-button>` without text and only icon, there is an extra margin to the right of the icon.

**To Reproduce**
1. Try this:
```
<go-button
          buttonIcon="delete"
          (handleClick)="random()"
        ></go-button>
```
2. Resultant is as per the attached screenshot.
![icon2](https://user-images.githubusercontent.com/3825309/83397113-b4f0b400-a41a-11ea-9d71-29d001a1f632.png)

**Expected behavior**
The DSM recommends a button like this:
![icon1](https://user-images.githubusercontent.com/3825309/83397193-d356af80-a41a-11ea-9848-0dc38f3b781f.png)

Ashygoyal:
Hi @grahamhency , 
There is a new issue as a result of the code changes wrt this issue.

If you notice the height of the button with text (and icon) and the height of the button with only icon (no text), there is a discrepancy.
![image](https://user-images.githubusercontent.com/3825309/84038477-9f046580-a9bd-11ea-98b9-d84db3da0a5e.png)

grahamhency:
@Ashygoyal on https://tangoe.design/ui-kit/button they are the same height. Can you confirm that the "hello" button here is a `go-button` and there are no custom styles being applied?

Ashygoyal:
Sure @grahamhency,
Sharing the code snippet below:
![image](https://user-images.githubusercontent.com/3825309/84046232-708b8800-a9c7-11ea-9575-139a4a665f76.png)

Also, I assure there is no forcefully applied custom css via ng-deep.

Ashygoyal:
Infact, I can confirm that even here: https://tangoe.design/ui-kit/button
They are still not the same height. Notice that extra part of the icon button below the dotted line.
![Screenshot (57)](https://user-images.githubusercontent.com/3825309/84046808-3078d500-a9c8-11ea-8334-e85bd788908f.png)

grahamhency:
@Ashygoyal thanks for confirming those. Which browser are you using? Because on my end in Chrome I see them as all the same height.

Ashygoyal:
I am on Chrome v83.0.4 (Windows 10).
Also confirmed the same issue on Chrome for MacOS.

grahamhency:
[...]

PR #589, opened 2020-06-16
Use css only to manage padding for button

eligolding:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/mobi/goponents/blob/master/CONTRIBUTING.md

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #563 

## What is the new behavior?
This is a css only approach to fixing #563. The current solution introduced in #568 uses ViewChild to check if the DOM node has text, and is being accessed in `ngOnInit` which does not work in Angular 9.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-button/go-button.component.scss
+++ b/projects/go-lib/src/lib/components/go-button/go-button.component.scss
@@ -44,6 +44,10 @@
   display: inline-flex;
 }
 
+.go-button--icon-only {
+  padding: calc(.625rem - 1px) .625rem;
+}
+
 .go-button__icon ~ .go-button__text:not(:empty) {
   padding-left: .5rem;
 }
--- a/projects/go-lib/src/lib/components/go-button/go-button.component.scss
+++ b/projects/go-lib/src/lib/components/go-button/go-button.component.scss
@@ -42,15 +42,10 @@
 
 .go-button__icon {
   display: inline-flex;
-  padding-right: .5rem;
 }
 
-.go-button--icon-only {
-  padding: calc(.625rem - 1px) .625rem;
-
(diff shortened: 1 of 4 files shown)


projects/go-lib/src/lib/components/go-button/go-button.component.scss
```diff
@@ -42,15 +42,10 @@
 
 .go-button__icon {
   display: inline-flex;
-  padding-right: .5rem;
 }
 
-.go-button--icon-only {
-  padding: calc(.625rem - 1px) .625rem;
```
grahamhency, line 49: We need this `.626rem` padding on the sides. It makes the icon only button oblong instead of a square:

Screen Shot 2020-06-17 at 12 50 57 PM

eligolding: added it back

projects/go-lib/src/lib/components/go-button/go-button.component.scss
```diff
@@ -42,15 +42,14 @@
...
 }
 
 .go-button--icon-only {
   padding: calc(.625rem - 1px) .625rem;
+}
 
-  .go-button__icon {
-    padding: 0;
-  }
+.go-button__icon ~ .go-button__text:not(:empty) {
```
grahamhency, line 51: This seems to have good browser support. The only caveat here is: https://caniuse.com/#feat=mdn-css_selectors_empty_matches_whitespace. Which we shouldn't really have an issue with, but if we do it's linked here for documentation.

Issue #17, opened 2019-04-16
<go-button> component doesn't have a focus style


AlexOverbeck:
The `<go-button>` components css hides the focus state of the buttons. This makes it harder to navigate via keyboard and overall less accessible.

https://github.com/mobi/goponents/blob/master/projects/go-lib/src/lib/components/go-button/go-button.component.scss

We should remove that style and make it more accessible. It also looks like this file doesn't quite follow BEM standards correctly. It might be a good opportunity to update some of those things as well.

AlexOverbeck:
This was fixed in #26

PR #596, opened 2020-06-22
[Chore] Refactor Branding Config

grahamhency:
## PR Checklist

## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The current implementation of the branding/config service is rather complex and intricate to set up. 

## What is the new behavior?
This refactor greatly simplifies the way the logo header area is configured as well as simplifies the way branding is applied throughout the app.

- Added the ability to specify a logo when the nav is collapsed
- Added `ThemeColors`, `BrandingMode`, and `LogoConfig` interfaces to `GoConfigInterface`
- Added the ability to `setConfig` on `GoConfigService`
- Added the ability to `getConfig` on `GoConfigService`
- Added the ability to `setLogo` on `GoConfigService`
- Added a new utility file with exports for:
   - `RGB` interface
   - `contrastIsAccessible` method
   - `shadeHex` method
   - `shadeRGB` method
   - `rgbToHex` method
   - `hexToRgb` method
   - `luminance` method
   - `calculateContrast` method

- [x] Yes
- [ ] No

### Breaking Changes

1. GoConfigService
   - The `config` observable has been renamed to `config$`
   - Removed the `toggleHeaderBrandingEnabled` method in favor of `setConfig`
   - Removed the `overrideMenuColor` method
   - Moved the `contrastIsAccessible` method to the new `colors.util.ts` file
   - Moved the `RGB` interface to the new `colors.util.ts` file
2. GoHeader Component
   - Removed bindings for `altText`, `logo`, `logoLink`, and `logoLinkUseRouterLink` in favor of setting them in the `setConfig` method on the `GoConfigService`
3. GoConfigInterface
   - Removed the `brandFontColor` property as it is no longer necessary
   - Removed the `headerBrandingEnabled` property as it has been replaced by the `brandingMode` property


projects/go-style-guide/src/app/features/ui-kit/components/configuration-docs/configuration-docs.component.html
```diff
@@ -2,6 +2,39 @@
...
+    class="go-column go-column--100">
+    <div class="go-container" go-card-content>
+      <div class="go-column go-column--100 go-column--no-padding">
+        <p class="go-body-copy">
+          The <code class="code-block--inline">GoConfigurationService</code> is used to apply a global set of configutions
+          for the app. Primarily this service is focused on branding configurations.
+        </p>
+        <p class="go-body-copy go-body-copy--no-margin">
+          The service uses a <code class="code-block--inline">GoConfigInterface</code> to store the configuration data. This
+          interface looked like this:
```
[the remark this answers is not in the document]
grahamhency: Yep, thanks.

.gitignore
```diff
@@ -38,3 +38,4 @@ twp.json
 # System Files
 .DS_Store
 Thumbs.db
+twp.json
```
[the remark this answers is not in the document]
grahamhency: Heh, this _probably_ doesn't need to be in this PR.. but it's for the Teamwork VSCode extension.

projects/go-lib/src/lib/go-config.service.spec.ts
```diff
@@ -1,12 +1,20 @@
 import { TestBed } from '@angular/core/testing';
 import { skip } from 'rxjs/operators';
-import { GoConfigInterface } from './go-config.model';
+import { BrandingMode, GoConfigInterface } from './go-config.model';
 import { GoConfigService } from './go-config.service';
 
 
-describe('GoConfigService', () => {
+fdescribe('GoConfigService', () => {
```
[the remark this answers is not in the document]
grahamhency: Oops. Thanks.

projects/go-lib/src/lib/go-config.service.ts
```diff
@@ -1,89 +1,62 @@
...
    */
-  public contrastIsAccessible(backgroundHex: string, foregroundHex: string): boolean {
-    const backgroundRgb: RGB = this.hexToRgb(backgroundHex);
-    const foregroundRgb: RGB = this.hexToRgb(foregroundHex);
-
-    const contrast: number = this.contrast(backgroundRgb, foregroundRgb);
-
-    return contrast > 4.5;
+  public setConfig(config: GoConfigInterface): void {
+    this.config = Object.assign({}, config);
```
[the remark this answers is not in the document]
grahamhency: Yeahhhhhh, good call. Really hope JavaScript figures out how to deep copy something at some point.

grahamhency:
@jaredami I fixed that failing test, would you mind giving another review?

projects/go-lib/src/lib/go-config.service.ts
```diff
@@ -1,89 +1,62 @@
...
-    brandColor: '#65B360',
-    headerBrandingEnabled: false
-  });
+  private config: GoConfigInterface = {
+    brandColor: ThemeColors.brand,
+    brandingMode: BrandingMode.company,
+    logoConfig: {
+      logo: '',
+      logoLink: ''
+    }
```
StevenUlmer, line 18: Should this be initialized without the `logoConfig` here? In the html for the header we have an `*ngIf="logoConfig"` and then we assume that if we have the logoConfig then we will have a logoLink, which won't be the case when this is initialized.

projects/go-lib/src/lib/go-config.model.ts
```diff
@@ -1,5 +1,78 @@
...
+   * Company mode is for branding the app in accordance
+   * to the constaints of the base company (ex. Tangoe)
+   */
+  company = 'Company',
+  /**
+   * Client mode is for branding the app in accordance
+   * to the standards for a client company (ex. External Branding / White-labelling)
+   */
+  client = 'Client'
+}
```
StevenUlmer, line 25: What is the actual difference between a company branded app and a client branded app here? The only thing that I see that uses this is the background of the header currently and even there it would seem that it's the difference between setting it to the company color, or setting it to white. I guess my question here is does Client Mode brand the app with External Branding as stated here, or is it just setting everything to the light theme?

projects/go-lib/src/lib/go-config.service.ts
```diff
@@ -1,89 +1,62 @@
...
+   * Configuration subject, you can subscribe to this to recieve change
+   * updates when the global config changes.
    */
-  public setBrandColor(color: string): void {
-    // we have to copy the config here or it won't regester a change in components
-    const config: GoConfigInterface = Object.assign({}, this.config.getValue());
-    config.brandColor = color;
-    this.config.next(config);
-  }
+  config$: BehaviorSubject<GoConfigInterface> = new BehaviorSubject<GoConfigInterface> (this.config);
```
StevenUlmer, line 25: So I know this is pretty standard to add a $ in angular to designate a variable that is an observable, but we haven't been following that practice anywhere else in goponents. This seems like a breaking change that could be avoided here by setting the private config to a variable called `_config` and keeping this as just config. I know we can't avoid all the breaking changes in this PR and some of them I think are necessary, but it still seems like we should avoid the ones we can. Thoughts?

projects/go-style-guide/src/app/features/ui-kit/components/configuration-docs/configuration-docs.component.html
```diff
@@ -54,81 +84,33 @@ <h3 class="go-heading-3">Update Branding Color</h3>
...
     </ng-container>
     <div class="go-container" go-card-content>
 
-      <div class="go-column go-column--100">
-        <p class="go-body-copy go-body-copy--no-margin">
-          The <code class="code-block--inline">public toggleHeaderBrandingEnabled(): void</code> method can 
-          be used to enable/disable header branding.
+      <div class="go-column go-column--50">
+        <p class="go-body-copy">
+          To enable the "Client Branding" mode, the config on the service must be reset.
```
StevenUlmer, line 93: Should we provide a method in the service to just set the `brandingMode`? We provide a method to set the `logo` and the `brandColor`, so for consistency sake should we also provide a way to change the branding mode with out resetting the entire config?

projects/go-lib/src/lib/go-config.model.ts
grahamhency: @StevenUlmer so there's an important distinction here between branding and theme. A theme would be light vs. dark. Branding is the color applied as an accent. In this particular case, the only main thing that this changes is how the branding color get's applied to the background of the header.

projects/go-lib/src/lib/go-config.service.ts
grahamhency: I suppose that makes sense. I was initially just trying to adopt the naming convention and figured that if we were already making a few breaking changes in this version, what's one more... but you're probably right, we should limit it if we can. I've changed things to match your suggestion here.

projects/go-lib/src/lib/go-config.service.ts
grahamhency: Sure, that shouldn't matter to remove this. :+1:

projects/go-style-guide/src/app/features/ui-kit/components/configuration-docs/configuration-docs.component.html
grahamhency: Sure, I think the use case for this will be much less because if you're changing the mode you're probably also changing the logos and brand color.. but it can't hurt to throw a couple lines of code in there just in case.

projects/go-lib/src/lib/components/go-side-nav/go-nav-item/go-nav-item.component.spec.ts
```diff
@@ -4,6 +4,9 @@ import { GoNavItemComponent } from './go-nav-item.component';
 import { GoIconModule } from '../../go-icon/go-icon.module';
 import { RouterTestingModule } from '@angular/router/testing';
 import { GoConfigService } from '../../../go-config.service';
+import { GoConfigInterface } from '../../../go-config.model';
+import { skip } from 'rxjs/operators';
```
jaredami, line 8: Is this used anywhere?

projects/go-lib/src/lib/go-config.model.ts
StevenUlmer: Ya I guess I was wondering if there might be a more descriptive variable name for these two variables based on what they do, as company vs client branding doesn't really tell me the difference, but honestly I can't think of a good name that is more descriptive so this is probably fine

projects/go-lib/src/lib/components/go-side-nav/go-nav-item/go-nav-item.component.spec.ts
grahamhency: Nope, removed it. Thanks.

projects/go-lib/src/lib/go-config.model.ts
Ashygoyal: Guys, this is super trivial, but check line #17, there is a typo: ~constaints~ constraints.
@StevenUlmer / @grahamhency

Issue #591, opened 2020-06-17
[BUG] Difference in heights of Input fields vs Select Dropdowns


Ashygoyal:
**Describe the bug**
 If you look at the input fields, they have a height of 36.67px.
But if you check the select dropdowns, their height is 36px.
(attaching the screenshots)

This difference is too small to notice, until you have a select and a go-button, side by side (because all our go-buttons also have a height of 36.67px).

My **system details**:
**OS**: Windows 10 Enterprise (x-64)
**Browser**: Chrome [Version 83.0.4103.97 (Official Build) (64-bit)]
**Screen Res**: 
![image](https://user-images.githubusercontent.com/3825309/84904835-6e16e580-b0cd-11ea-80a0-b262371ebc23.png)

**Expected behavior**
Input fields, select dropdowns, all types of buttons, etc should have the same height.

**Screenshots**
Screenshots taken at 100% browser zoom (default viewport).

![image](https://user-images.githubusercontent.com/3825309/84905041-bd5d1600-b0cd-11ea-86ab-04ef3bf1f7b6.png)
![image](https://user-images.githubusercontent.com/3825309/84905079-ccdc5f00-b0cd-11ea-901b-833522aa98f7.png)

PR #598, opened 2020-06-22
[Bug] Align Form Heights

grahamhency:
## PR Checklist

## PR Type
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: CSS Changes

The selects, buttons, and form inputs were inconsistent in heights
Issue Number: #591 

## What is the new behavior?
Aligned heights, font size, and line height for selects, buttons, and form inputs to match specifications in dsm

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-select/go-select.component.scss
+++ b/projects/go-lib/src/lib/components/go-select/go-select.component.scss
@@ -12,24 +12,33 @@
     font-size: .875rem;
     font-weight: $weight-regular;
     letter-spacing: .5pt;
+    line-height: 1.3;
+    padding: 0;
   }
 }
 
 .ng-select.ng-select-single .ng-select-container,
 .ng-select.ng-select-multiple .ng-select-container {
-  height: 2.25rem;
+  height: auto;
+  min-height: auto;
+}
+
+.ng-select.ng-select.ng-select-multiple .ng-select-container .ng-value-container,
+.ng-select.ng-select-single .ng-select-container .ng-value-container {
+  align-items: flex-start;
+  padding: 0;
 }
 
 // overrides position of input element
 .ng-select.ng-select-multiple .ng-select-container .ng-value-container,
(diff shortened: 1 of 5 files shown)


projects/go-lib/src/lib/components/go-select/go-select.component.scss
```diff
@@ -12,24 +12,33 @@
...
     font-weight: $weight-regular;
     letter-spacing: .5pt;
+    line-height: 1.3;
+    padding: 0;
   }
 }
 
 .ng-select.ng-select-single .ng-select-container,
 .ng-select.ng-select-multiple .ng-select-container {
-  height: 2.25rem;
```
StevenUlmer, line 20: Is there a reason for removing this height here? If we keep it we can avoid adding the `position: relative !important;` below I believe.

grahamhency: @StevenUlmer yes. This is being removed because we should not be specifying an exact height for this. The size of the element should depend upon the size of the font, line-height, and padding. If we were to change any of those it would affect the height. That's kind of why this PR exists.

Unfortunately `ng-select` applies styles directly to this element when it handles opening/closing which is why that `!important` is necessary.

projects/go-lib/src/lib/components/go-select/go-select.component.scss
```diff
@@ -47,7 +56,17 @@
...
   font-weight: $weight-light;
-  top: .75rem;
+  line-height: 1.3;
+  padding: .5rem;
+  position: absolute;
+  top: 0;
+}
+
+.ng-select.ng-select-single .ng-select-container .ng-value-container .ng-value {
+  line-height: 1.5;
```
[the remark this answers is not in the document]
grahamhency: Bummer, nope. I switched them to `1.3`

PR #599, opened 2020-06-23
[Feature] Table Title Template

### LINKED ISSUE #519: [Feature] Custom Table Headers
We would like the ability to customize table headers a bit more.  

_**Example:**_
We have a need for some table headers to be go-heading-4s  instead of go-heading-3s since these tables appear within cards.

We also have a need to have the table headers function as a ng-select menu where the user can toggle  through multiple tables in a single accordion.

We would like the ability in go-ponents to customize the table-header element a bit more. These customizations include replacing the title with a ng-select menu or changing the header size.

_**Example:**_
You can see an example of the select menu being used as a table-header in the following example:
https://tangoe.invisionapp.com/share/URUHTZLWFQS#/396747554_Invoice_Detail

Scroll down to View By and you can see an example of this.

grahamhency:
## PR Checklist

## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The table title can only be a string passed in as a binding through the table.
Resolves #519 

## What is the new behavior?
An `ng-template` can now be declared in place of the `tableTitle` area.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.ts
@@ -66,6 +66,7 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
 
   @ContentChildren(GoTableColumnComponent) columns: QueryList<GoTableColumnComponent>;
   @ContentChild('goTableDetails', { static: false }) details: TemplateRef<any>;
+  @ContentChild('goTableTitle', { static: false }) tableTitleTemplate: TemplateRef<any>;
 
   @ViewChild('selectAllCheckbox', { static: false }) selectAllCheckbox: ElementRef;
 
--- a/projects/go-style-guide/src/app/features/ui-kit/components/table-docs/components/table-title-template/table-title-template.component.html
+++ b/projects/go-style-guide/src/app/features/ui-kit/components/table-docs/components/table-title-template/table-title-template.component.html
@@ -1,0 +1,41 @@
+<section class="go-container">
+
+  <go-card class="go-column go-column--100">
+    <ng-container go-card-header>
+      <h2 class="go-heading-2">Implementing a Title Template</h2>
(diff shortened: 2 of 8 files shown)


projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -66,6 +66,7 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit {
 
   @ContentChildren(GoTableColumnComponent) columns: QueryList<GoTableColumnComponent>;
   @ContentChild('goTableDetails', { static: false }) details: TemplateRef<any>;
+  @ContentChild('goTableTitle', { static: false }) tableTitleTemplate: TemplateRef<any>;
```
StevenUlmer, line 69: I feel like the title should be static, but I could maybe see the case where an icon might change or something

projects/go-style-guide/src/app/features/ui-kit/components/table-docs/components/table-title-template/table-title-template.component.html
```diff
@@ -0,0 +1,41 @@
...
+    </ng-container>
+  </go-card>
+
+  <go-table
+    class="go-column go-column--100"
+    [tableConfig]="tableConfig">
+    <ng-template #goTableTitle>
+      <go-button (handleClick)="makeToast()">
+        Make Toast
+      </go-button>
```
StevenUlmer, line 31: Should we maybe make this a little more realistic of an example? This seems like a good test case to make sure it works, but I don't see this actually being used like this. Maybe make the example be a header with a go-icon-button beside it that when clicked makes a toast or something?

grahamhency: That's fair, I've updated.

PR #642, opened 2020-08-17
Update table design

Ashygoyal:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our [guidelines](https://github.com/mobi/goponents/blob/master/CONTRIBUTING.md)
- [ ] Tests for the changes have been added
- [ ] Docs have been added or updated

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [x] Style Update (CSS)
- [ ] Other... Please describe:

## What is the current behavior?

Resolves [605](https://github.com/mobi/goponents/issues/605)

## What is the new behavior?

Styling changes for table action icons, table column header and table search input.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #641, opened 2020-08-14
Large off canvas

eligolding:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our [guidelines](https://github.com/mobi/goponents/blob/master/CONTRIBUTING.md)
- [ ] Tests for the changes have been added
- [ ] Docs have been added or updated

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Style Update (CSS)
- [ ] Other... Please describe:

## What is the current behavior?

Resolves 

## What is the new behavior?
An option to have the off canvas to overlay 75% of the page, as in the [Tangoe DSM](https://tangoe.invisionapp.com/dsm/tangoe/tangoe-ds/folder/components/5e2203660dc888b5b891e67c)

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #647, opened 2020-08-25
v1.6.1

grahamhency:
###### Change log for `v1.6.1`

***

# 🚢  New Features

### Off Canvas
- New bindings and features - #641 
   - Ability to change size of off canvas
   - Ability to change theme of off canvas
   - Ability to pass content into header of off canvas

### Portal
- New component/directive used to render content from one component elsewhere in the app - #641 

# 🐞 Bug Fixes

### Button
- Fixed an issue where the line height of the button was incorrect - #637 

### Off Canvas & Modal
- Fixed an issue where dynamically created components were never destroyed - #645 

# ✏️ General & Design

### Design
- Updated table design - #642 

### Misc
- Added modifier class to grid column to allow column to expand to 100% width - #638 
- Corrected minor issues with documentation - #643


eligolding:
@StevenUlmer making this full width would be a bit more involved, so in the meantime I put up a [PR](#650) to have the header content wrap so you don't have the overflow that you were seeing.

PR #692, opened 2020-10-26
Emit rowSelectionEvent when selecting or deselecting all table rows

### LINKED ISSUE #684: [Chore] Emit rowSelectionEvent when selecting/deselecting all rows in go-table
Currently, we emit an event from `rowSelectionEvent` (which includes `selectedRows`) on the `go-table` anytime a table row is selected/deselected. 

However, we have come across at least one use-case where we want to emit from `rowSelectionEvent` when the checkbox to select/deselect *all* table rows is clicked in addition to doing so when individual rows are selected, and currently we do not emit an event from `rowSelectionEvent` when selecting/deselecting all rows. We should add this functionality.

jaredami:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)
- [x] Other... Please describe:

Resolves #684 

## What is the new behavior?
The `go-table` now emits an event from the `rowSelectionEvent` output when the checkbox to select/deselect all rows is clicked, rather than just when individual rows are selected/deselected.

This will enable components that need to be aware of the table's selection state when the checkbox to select/deselect all rows is clicked to do so.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
@@ -2,7 +2,7 @@ export interface RowSelectionEvent extends SelectionState {
   /**
    * The current row that was targeted for selection
    */
-  currentRow?: {
+  currentRow: {
     /**
      * The entire data in the row
      */
--- a/projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table-selection.model.ts
@@ -2,7 +2,7 @@ export interface RowSelectionEvent extends SelectionState {
   /**
    * The current row that was targeted for selection
    */
-  currentRow: {
+  currentRow?: {
     /**
      * The entire data in the row
      */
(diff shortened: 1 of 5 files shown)


grahamhency (changes requested):
@jaredami I have some concerns about what kind of side effects introducing this change might have with current implementations watching for the `rowSelectionEvent` already. That event also is named specific to a row, even though it provides the entire selection state.

What do you think about introducing a new event (let's say `selectAllEvent`) that only fires when the state of the `selectAll` checkbox state changes? It would allow us to handle the functionality differently on implementation, removes the potential for unintended side effects of firing the `rowSelectionEvent` an additional time, at the expense of having to handle another event.

What're your thoughts?

jaredami:
@grahamhency  Those are good points. I didn't realize that this could potentially constitute a breaking change.

I went ahead and did something like you suggested. The trickiest part was updating the docs. Let me know if you think those are clear enough.

grahamhency (approved):
I'm good with this approach, I think it's safer and the documentation looks good to me. :+1:

PR #693, opened 2020-10-27
Unsubscribe in components to prevent memory leaks

### LINKED ISSUE #689: [Chore] Audit components for missing unsubscribes
We should go through and make sure there are not any instances where we are failing to unsubscribe from (non-http) subscriptions in the `ngOnDestory` of components. This will ensure we do not have issues with data leaks.

jaredami:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)
- [x] Other... Please describe:

Resolves #689 

## What is the new behavior?
This ensures that we are unsubscribing from all of the non-http subscriptions within our components in order to prevent potential memory leaks.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.ts
@@ -15,8 +15,8 @@ import {
   TemplateRef
 } from '@angular/core';
 import { FormBuilder, FormControl, FormGroup } from '@angular/forms';
-import { Subscription } from 'rxjs';
-import { debounceTime, distinctUntilChanged } from 'rxjs/operators';
+import { Subject } from 'rxjs';
+import { debounceTime, distinctUntilChanged, takeUntil } from 'rxjs/operators';
 import { fadeTemplateAnimation } from '../../animations/fade.animation';
 import { detailButtonAnim, tableRowBorderAnim } from '../../animations/table-details.animation';
 import { GoTableColumnComponent } from './go-table-column.component';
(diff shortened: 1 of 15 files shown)


StevenUlmer (dismissed):
I think this all looks good. The one thing I think might be worth taking a look into is creating a BaseComponent class that all of these other components extend from that has the logic for the `ngOnDestroy` in it. That way for any other components that come along where we need to unsubscribe we can just extend that BaseComponent and we won't have to add this logic again. I'll leave that up to you though.

jaredami:
>The one thing I think might be worth taking a look into is creating a BaseComponent class that all of these other components extend from that has the logic for the `ngOnDestroy` in it. That way for any other components that come along where we need to unsubscribe we can just extend that BaseComponent and we won't have to add this logic again.

@StevenUlmer I'm thinking it might be better to leave things more "exposed" so that you don't have to dig down to see what is going on and to understand why we are doing it this way, if that makes sense.

It is requiring another review after rebasing though, when you get a chance 👍

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -490,8 +496,7 @@ export class GoTableComponent implements OnInit, OnChanges, AfterViewInit, OnDes
   }
 
   private setupRowSelectFormForCurrentPage(): void {
-    this.rowControlSubsCollection.unsubscribe();
-    this.rowControlSubsCollection = new Subscription();
+    this.pageChange$.next();
```
StevenUlmer, line 499: Does calling `this.pageChange$.next();` here actually cleans up the subscription? I believe we would still either need to call `.complete()` or `.unsubscribe()` here and then set `this.pageChange$ = new Subject()` in order to actually clean up this subscription

StevenUlmer (changes requested):
Just the one question related to the new changes

projects/go-lib/src/lib/components/go-table/go-table.component.ts
jaredami: I think we're good here because of this part:

```javascript
  private setupRowSelectFormControlsSubs(): void {
    this.getDisplayData().forEach((row: any) => {
      this.rowSelectForm.get(`selection_${row[this.localTableConfig.selectBy]}`)
        .valueChanges.pipe(takeUntil(this.pageChange$))
        .subscribe((value: boolean) => {
          this.selectionChange(value, row);
        });
    });
  }
```

Here we're subscribing to each control, but then each of those subscriptions is watching `this.pageChanges$` and unsubscribing from the control when `this.pageChanges$` emits. 

Here's what happens if I log `pageChange$` every time the page changes and then switch pages a couple times:

Screen Shot 2020-11-04 at 3 55 28 PM

If it was duplicating the subscriptions, I think you'd expect there to be 10, then 20, and then 20 `observers`.

So it's like the `pageChanges$` gets "emptied out" every time the page changes due to the `takeUntil`s and then is ready to receive a new set of observers. Does that make sense or am I missing something?

PR #701, opened 2020-10-29
Add closeOnClick input to go-panel

### LINKED ISSUE #700: [Chore] Add input to determine whether clicking a go-panel closes the action sheet
Currently, clicking on a `go-panel` within a `go-action-sheet` always closes the action sheet. We have run in to some instances where we do want this to happen. We should add an input to the `go-panel` to enable making it so the action sheet does *not* close when the panel is clicked.

jaredami:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

Resolves #700 

## What is the new behavior?
This adds a `closeOnClick` input to the `go-panel` which determines whether clicking the panel closes its parent action-sheet. The input is set to `true` by default.

This was needed because there are some cases where we do not want action-sheets to close when the `go-panel`s within them are clicked.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.spec.ts
+++ b/projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.spec.ts
@@ -44,22 +44,31 @@ describe('GoPanelComponent', () => {
       expect(component.action.emit).toHaveBeenCalled();
     });
 
-    it('should close the action sheet', () => {
-      parentComponent.showContent = true;
+    it('should call panelClicked', () => {
+      spyOn(component, 'panelClicked');
 
       component.handleAction();
 
-      expect(parentComponent.showContent).toBeFalsy();
+      expect(component.panelClicked).toHaveBeenCalled();
     });
   });
 
-  describe('closeActionSheet', () => {
-    it('should set parent action sheet\'s showContent to false', () => {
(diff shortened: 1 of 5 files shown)


projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.spec.ts
```diff
@@ -44,22 +44,31 @@ describe('GoPanelComponent', () => {
...
-    it('should set parent action sheet\'s showContent to false', () => {
+  describe('panelClicked', () => {
+    it('should set parent action sheet\'s showContent to false if closeOnClick is true', () => {
       parentComponent.showContent = true;
 
-      component.closeActionSheet();
+      component.panelClicked();
 
       expect(parentComponent.showContent).toBeFalsy();
     });
```
StevenUlmer, line 63: Should we specifically set `component.closeOnClick = true` here since it doesn't look like were doing any clean up after these tests which means if at any point the order of these tests got changed this test would fail?

StevenUlmer (changes requested):
Just the one comment. Everything else looks good.

StevenUlmer (approved):
Looks good to me

PR #699, opened 2020-10-29
Max and min length bindings for goInput - 691

Ashygoyal:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our [guidelines](https://github.com/mobi/goponents/blob/master/CONTRIBUTING.md)
- [ ] Tests for the changes have been added
- [x] Docs have been added or updated

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [ ] Style Update (CSS)
- [ ] Other... Please describe:

## What is the current behavior?
Right now, goponents does not provide a way to set maxlength and minlength for `go-input`. This PR includes necessary changes (both code-wise and documentation) for the same.

Resolves #691 

## What is the new behavior?
Developers can now set the maxlength and minlength for `go-input`

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #706, opened 2020-10-30
Text Area Max Length

### LINKED ISSUE #705: [Feature] Max Length for Text Area
I am requesting that we expose the `maxlength` attribute for the `text area`. There is already an issue for exposing the maxlength attribute on inputs, but since the text-area is a different component we will also need to expose it there as well

On a `go-text-area`, be able to pass a value `maxlength` and have it apply to the `text-area` element inside the `go-text-area`.

StevenUlmer:
## PR Checklist

## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

Resolves #705 

## What is the new behavior?
There is the ability to pass in a maxlength on the text area

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
+++ b/projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
@@ -16,6 +16,7 @@
     [rows]="rows"
     [formControl]="control"
     [attr.maxlength]="maxlength || null"
+    [attr.minlength]="minlength || null"
   ></textarea>
 
   <go-hint *ngFor="let hint of hints" [message]="hint" [theme]="theme"></go-hint>
--- a/projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
+++ b/projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
@@ -15,6 +15,7 @@
     [placeholder]="placeholder"
     [rows]="rows"
     [formControl]="control"
+    [attr.maxlength]="maxlength || null"
   ></textarea>
 
   <go-hint *ngFor="let hint of hints" [message]="hint" [theme]="theme"></go-hint>
--- a/projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
+++ b/projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
@@ -15,8 +15,8 @@
     [placeholder]="placeholder"
     [rows]="rows"
     [formControl]="control"
-    [attr.maxlength]="maxlength || null"
-    [attr.minlength]="minlength || null"
+    [maxlength]="maxlength"
+    [minlength]="minlength"
   ></textarea>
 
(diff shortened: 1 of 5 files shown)


projects/go-style-guide/src/app/features/ui-kit/components/form-docs/components/text-area-docs/text-area-docs.component.html
```diff
@@ -345,4 +345,42 @@ <h4 class="go-heading-4 go-heading--underlined">Code</h4>
...
+  <go-card id="form-textarea-max-length" class="go-column go-column--100">
+    <ng-container go-card-header>
+      <h2 class="go-heading-2 go-heading--no-wrap">Component Max Length</h2>
+      <go-copy cardId="form-textarea-max-length" goCopyCardLink></go-copy>
+    </ng-container>
+    <div class="go-container" go-card-content>
+
+      <div class="go-column go-column--100">
+        <p class="go-body-copy go-body-copy--no-margin">
+          The text area component's max legnth can be set via
```
jaredami, line 358: Should "legnth" be "length"?

jaredami (changes requested):
Just noticed a typo but looks good otherwise 👍

grahamhency:
@StevenUlmer, what if we also add on the `minlength` attribute to this? Probably not going to be used very often, but we're doing it in this PR as well #699

grahamhency (dismissed):
Looks good, thanks!

projects/go-lib/src/lib/components/go-text-area/go-text-area.component.html
```diff
@@ -15,6 +15,8 @@
     [placeholder]="placeholder"
     [rows]="rows"
     [formControl]="control"
+    [attr.maxlength]="maxlength || null"
+    [attr.minlength]="minlength || null"
```
jaredami, line 19: One question I have about this is whether we should be using `attr.` for these. In the [Angular docs](https://angular.io/guide/binding-syntax) it says: "It is recommended that you set an element property with a property binding whenever possible."

Also [here](https://angular.io/guide/binding-syntax) it says "Attributes initialize DOM properties and then they are done. Property values can change; attribute values can't." I'm wondering if that means that when you use `attr.` it makes the binding static (i.e. it won't update if the value is changed)?

These two things make me think maybe we should just property binding here, unless the `text-area`s don't have these properties.

jaredami (changes requested):
Couple of things in addition to my question about attribute vs property binding:

1. Should we add default values and add some logic to ensure that the `minlength` can't be set to greater than the `maxlength`, like what was done [here](https://github.com/mobi/goponents/pull/699#event-3975155105)?
2. Should we *technically* use brackets when setting the binding values since the inputs are of type `number` rather than `string`? Maybe not worth caring about...

grahamhency (approved):

PR #714, opened 2020-11-11
[Bug] Enforce Table Column Width

### LINKED ISSUE #707: [BUG] Go-table (min-width) column Issue
Unable to increase the width of the column in the go-table, the column stays the same width as the column header after increasing the width to 400.

**To Reproduce**
1. Add 17 or more columns inside the go-table.
2. try to increase the width of one of the columns to 400.
3. save the changes and check if the width is changing.

**Expected behavior**
require the increase the width of the column body more than the column header width.

grahamhency:
## PR Checklist
~~Tests for the changes have been added~~

## PR Type
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

Currently we set the `width` property on the table cell when assigned one. That approach does not constrain the width of the column to the desired size if the content within the table is too great for its container.

Resolves #707 

## What is the new behavior?
Now, the `min-width` property is set instead of the `width`. This allows for stricter control over table column widths. Using `min-width` will allow the column to expand if the space is available, but it will not shrink beyond the specified limit regardless of other table content.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.html
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.html
@@ -59,7 +59,8 @@
           <th *ngFor="let col of columns"
               class="go-table__head-column"
               (click)="toggleSort(col.sortable, col.field)"
-              [ngClass]="{ 'go-table__head--sortable': col.sortable !== undefined ? col.sortable : localTableConfig.sortable }">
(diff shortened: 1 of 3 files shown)


jaredami (dismissed):
Looks good to me. Just had one thought: should we maybe explain in the docs that the `width` input just sets the *minimum* width?

Issue #717, opened 2020-11-19
[BUG]  Datepicker calendar hidden inside of modal.


amitpatra91:
**Describe the bug**
When I will use go-datepicker inside of the modal. And if I will click that input filed then the calendar is not displaying properly. It is hiding inside of the modal.

**Expected behavior**
So that calendar should display properly, it should not hide inside of go-modal.

**Screenshots**
I have attached the screenshot please check that attached screen for a batter understanding

![MicrosoftTeams-image (5)](https://user-images.githubusercontent.com/33975358/99653846-8fff0480-2a7f-11eb-8c42-9d89a5ba3e66.png)

PR #720, opened 2020-11-26
Add appendTo features of go-datepicker module

amitpatra91:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

After clicking on the calendar icon calendar body was hiding behind of modal body.

Resolves [717](https://github.com/mobi/goponents/issues/717)

## What is the new behavior?
Now it's displayed on the modal body.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.html
+++ b/projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.html
@@ -1,9 +1,8 @@
 <div class="go-calendar--wrapper">
   <div
     class="go-calendar"
-    [ngClass]="{'go-calendar--above': displayAbove, 'go-calendar--right': displayFromRight, 'go-calendar--append-to-content': appendToContent}"
-    #calendarView
-    >
+    [ngClass]="{ 'go-calendar--above': displayAbove, 'go-calendar--right': displayFromRight, 'go-calendar--append-to-content': appendToContent }"
+    #calendarView>
     <div *ngIf="view === 'day'">
       <go-calendar-day-view
         [day]="selectedDate"
--- a/projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.html
+++ b/projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.html
@@ -1,6 +1,6 @@
 <div class="go-calendar--wrapper">
   <div
-    class="go-calendar"
+    class="go-calendar append-body"
     [ngClass]="{'go-calendar--above': displayAbove, 'go-calendar--right': displayFromRight, 'go-calendar--append-to-content': appendToContent}"
     *ngIf="opened">
     <div *ngIf="view === 'day'">
--- a/projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.html
+++ b/projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.html
@@ -1,8 +1,10 @@
 <div class="go-calendar--wrapper">
   <div
-    class="go-calendar append-body"
+    class="go-calendar"
     [ngClass]="{'go-calendar--above': displayAbove, 'go-calendar--right': displayFromRight, 'go-calendar--append-to-content': appendToContent}"
-    *ngIf="opened">
+    *ngIf="opened"
+    #calenderView
+    >
     <div *ngIf="view === 'day'">
       <go-calendar-day-view
         [day]="selectedDate"
--- a/projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.html
+++ b/projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.html
@@ -2,8 +2,7 @@
   <div
     class="go-calendar"
     [ngClass]="{'go-calendar--above': displayAbove, 'go-calendar--right': displayFromRight, 'go-calendar--append-to-content': appendToContent}"
-    *ngIf="opened"
-    #calenderView
+    #calendarView
     >
     <div *ngIf="view === 'day'">
       <go-calendar-day-view
--- a/projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts
+++ b/projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts
@@ -1,4 +1,15 @@
-import { Component, ElementRef, EventEmitter, HostListener, Input, OnDestroy, OnInit, Output, ViewChild } from '@angular/core';
+import {
+  AfterViewChecked,
+  Component,
+  ElementRef,
+  EventEmitter,
+  HostListener,
+  Input,
+  OnDestroy,
+  OnInit,
+  Output,
+  ViewChild
+} from '@angular/core';
 import { fadeAnimation } from '../../animations/fade.animation';
 import { GoCalendar } from './go-calendar';
 import { DateAdapter } from './date-adapter';
@@ -12,7 +23,7 @@ import { CalendarCell } from './calendar-cell.model';
     fadeAnimation
   ]
 })
-export class GoCalendarComponent implements OnDestroy, OnInit {
+export class GoCalendarComponent implements OnDestroy, OnInit, AfterViewChecked {
   canClose: boolean = true;
   currentMonth: number = 0;
   currentYear: CalendarCell;
@@ -32,7 +43,7 @@ export class GoCalendarComponent implements OnDestroy, OnInit {
   @Input() appendTo: string;
 
   @Output() datePicked: EventEmitter<Date> = new EventEmitter<Date>();
-  @ViewChild('calenderView', { static: false }) calenderView: ElementRef;
+  @ViewChild('calendarView', { static: true }) calendarView: ElementRef;
 
   constructor() {
     this.dateAdapter = new DateAdapter();
@@ -51,11 +62,6 @@ export class GoCalendarComponent implements OnDestroy, OnInit {
     this.canClose = false;
   }
 
-  @HostListener('document:mousedown', ['$event'])
-  onClick(event: MouseEvent): void {
-    this.removeCalenderfromBody();
-  }
-
   @HostListener('window:keydown', ['$event'])
   keyEvent(event: KeyboardEvent): void {
     switch (event.key) {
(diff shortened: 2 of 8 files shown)


projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts
```diff
@@ -121,5 +122,25 @@ export class GoCalendarComponent implements OnDestroy, OnInit {
     }
     this.currentMonth = this.selectedDate.getMonth();
     this.updateYear(this.selectedDate.getFullYear());
+    if (this.appendTo === 'body') {
+      setTimeout(() => {
+        const calenderBody: any = document.getElementsByClassName(
+          'append-body'
+        )[0];
```
grahamhency, line 129: This will break if we have an instance with multiple datepickers on a page. What if we changed this to an Angular `ViewChild` reference instead?

projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts
```diff
@@ -121,5 +122,25 @@ export class GoCalendarComponent implements OnDestroy, OnInit {
     }
     this.currentMonth = this.selectedDate.getMonth();
     this.updateYear(this.selectedDate.getFullYear());
+    if (this.appendTo === 'body') {
+      setTimeout(() => {
```
grahamhency, line 126: Does this need to be in a `setTimeout`?

projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts
```diff
@@ -121,5 +122,25 @@ export class GoCalendarComponent implements OnDestroy, OnInit {
...
+        const calenderBody: any = document.getElementsByClassName(
+          'append-body'
+        )[0];
+        const calenderBodyPosition: any = document
+          .querySelector('.append-body')
```
grahamhency, line 131: If we change the above to be a `ViewChild` reference we can remove this `querySelector` and call `getBoundingClientRect` on the `ViewChild` element.

projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts
```diff
@@ -121,5 +122,25 @@ export class GoCalendarComponent implements OnDestroy, OnInit {
...
+          .getBoundingClientRect();
+        Object.assign(calenderBody.style, {
+          height: `${calenderBodyPosition.height}px`,
+          bottom: `${calenderBodyPosition.bottom}px`,
+          top: `${calenderBodyPosition.top}px`,
+          right: `${calenderBodyPosition.right}px`,
+          width: `${calenderBodyPosition.width}px`,
+          left: `${calenderBodyPosition.left}px`,
+          visibility: 'visible',
+        });
```
grahamhency, line 141: Do we have to assign all of these properties when doing this? Specifically `height`/`width`?

projects/go-lib/src/lib/components/go-datepicker/go-datepicker.component.ts
```diff
@@ -29,6 +29,7 @@ export class GoDatepickerComponent implements OnDestroy, OnInit {
   @Input() minDate: Date | string;
   @Input() placeholder: string = '';
   @Input() theme: string = 'light';
+  @Input() appendTo: string = 'default';
```
grahamhency, line 32: I think we should specify which string values are valid here. 
```ts
@Input() appendTo: string = 'body' | null;

File: projects/go-style-guide/src/app/features/ui-kit/components/form-docs/components/datepicker-docs/datepicker-docs.component.ts [hunk 6]
```diff
@@ -151,7 +152,23 @@ export class DatepickerDocsComponent implements OnInit {
...
+        placeholder: '10/28/1999',
+        label: 'Date Of Birth',
+      }
+    );
+  }
+  `;
+
+  constructor(
+    private subNavService: SubNavService,
+    private goModalService: GoModalService) {
```
grahamhency, line 171: Can you move these to a new line for cleanliness?
```ts
  private goModalService: GoModalService
) {

**Review by grahamhency — CHANGES_REQUESTED** [review]
This is a good first pass, I think we can make some improvements and simplifications though. Please see my comments.

Also, I looked at the deploy preview for this to see the functionality and there is a bug. To reproduce:
1. Go to the example in the docs and open the modal
2. Open the datepicker
3. Click the close icon for the modal
4. The datepicker remains on the page

Screenshot:
Screen Shot 2020-12-07 at 12 24 51 PM

File: projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts (hunk 4 above)
Comment (amitpatra91) [reply projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts:141]: If I have removed few properties so some alignment issue is coming. For that reason, I have added those properties. Can you please suggest to me if anything is not required. So I will remove those.

File: projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts (hunk 2 above)
Comment (amitpatra91) [reply projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts:126]: I think this setTimeout is required otherwise viewchild property is coming as an **undefined**. Because viewChild will work after dom load. For that reason, I have added that setTimeout.

File: projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts (hunk 1 above)
Comment (amitpatra91) [reply projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts:129]: Yes, you are right. I have removed and update the code here.

File: projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts [hunk 7]
```diff
@@ -63,15 +70,11 @@ export class GoCalendarComponent implements OnDestroy, OnInit {
...
-      if (value) {
         this.selectedDate = this.calendar.selectedDate;
         this.initializeDate();
-      }
-      this.opened = value;
-    });
-
-    this.dateAdapter.setLocale(this.locale);
+        this.appendToCalender();
+        this.opened = true;
```
grahamhency, line 76: We can actually remove the `opened` property altogether because we're checking the opened status with the `ngIf` on the `go-calendar` component itself in the datepicker component.

projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts
```diff
@@ -63,15 +70,11 @@ export class GoCalendarComponent implements OnDestroy, OnInit {
...
-    this.subscription = this.calendar.calendarOpen.subscribe((value: boolean) => {
-      if (value) {
         this.selectedDate = this.calendar.selectedDate;
         this.initializeDate();
-      }
-      this.opened = value;
-    });
-
-    this.dateAdapter.setLocale(this.locale);
+        this.appendToCalender();
```
grahamhency, line 75: We should move this method call into an `AfterViewChecked` call like so:
```ts
ngAfterViewChecked(): void {
  this.appendToCalender();
}
```

You'll need to add `AfterViewChecked` to the class implementation list. [See Angular's docs for an example](https://angular.io/api/core/AfterViewChecked#usage-notes)

projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts
```diff
@@ -63,15 +70,11 @@ export class GoCalendarComponent implements OnDestroy, OnInit {
...
+        this.opened = true;
+        this.dateAdapter.setLocale(this.locale);
   }
 
   ngOnDestroy(): void {
```
grahamhency, line 80: In this method, you'll want to add this method call above `this.closeCalendar();`. New method should look like:
```ts
ngOnDestroy(): void {
  this.removeCalenderfromBody();
  this.closeCalendar();
}
```

projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.ts
grahamhency: I've added some further comments, but this `setTimeout` can be removed with the `ViewChild` approach. Review my other comments, you'll need to make a couple adjustments to get it working properly.

projects/go-lib/src/lib/components/go-datepicker/go-datepicker.component.html
```diff
@@ -25,7 +25,9 @@
     (click)="toggleDatepicker($event)"
   ></go-icon-button>
   <go-calendar
+    *ngIf=goCalendar.isOpen
```
grahamhency, line 28: This should be `*ngIf="goCalendar.isOpen"`, notice the quotes.

projects/go-lib/src/lib/components/go-datepicker/go-calendar.component.html
```diff
@@ -2,7 +2,9 @@
   <div
     class="go-calendar"
     [ngClass]="{'go-calendar--above': displayAbove, 'go-calendar--right': displayFromRight, 'go-calendar--append-to-content': appendToContent}"
-    *ngIf="opened">
+    *ngIf="opened"
```
grahamhency, line 5: This line can be removed, along with the `opened` property.

grahamhency (changes requested):
Sorry for the delay, had to do some research and some trial & error to figure out how to not use the `setTimeout` and other things. I've made some more comments, let me know your thoughts.

projects/go-lib/src/lib/components/go-datepicker/go-datepicker.component.ts
```diff
@@ -29,6 +29,7 @@ export class GoDatepickerComponent implements OnDestroy, OnInit {
   @Input() minDate: Date | string;
   @Input() placeholder: string = '';
   @Input() theme: string = 'light';
+  @Input() appendTo: string = null;
```
grahamhency, line 32: Just to clarify, I still think we should type this as:
```ts
@Input() appendTo: 'body' | null = null;

grahamhency (changes requested):
Just one more comment on types.

Additionally, I think we need to write some tests for these changes and also fix the broken tests.

grahamhency (changes requested):
Code is good, but there are broken tests and we need tests for the updated code.

grahamhency (changes requested):
The changes you proposed for the tests is not a solution. Those tests should still be in place an pass successfully.

grahamhency (approved):

PR #723, opened 2020-12-08
[Chore] Modal & Off Canvas Type Improvements

grahamhency:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

This PR addresses the potential for mismatched typing for both the off-canvas and modal implementations.

#### Modal
Currently the `openModal` method allows for `any` type to be passed in for the `component` and `bindings` properties. Additionally, the bindings for the modal itself were passed in through the `bindings` object which was originally intended to be for the bindings on the component being rendered by the modal.

#### Off Canvas
Currently the `openOffCanvas` method allows for `any` type to be passed in for the `component` and `bindings` properties. Additionally the bindings for the off canvas are set through properties on the object passed into this method.

#### Modal
There is now partial type enforcement on the `bindings` param. It will only allow for properties that exist on the Component passed into the `component` param or the properties available on the modal itself (although this is deprecated and should not be used in new code).

All modal implementations that require properties on the modal itself to be set should now use the `modalOptions` optional param available on the `openModal` method.

#### Off Canvas
There is now partial type enforcement on the `bindings` property. It will only allow for properties that exist on the Component passed into the `component` property.

The existing properties for `header` and `size` remain functional, but are now deprecated and should not be used.

All off canvas implementations that require properties on the off canvas itself to be set should now use the `offCanvasOptions` optional property.

- [ ] Yes
- [x] No

This PR was engineered in a way such that it will not introduce breaking changes into your app (unless the modal or off-canvas implementations were done incorrectly). We did include deprecation warnings where applicable to help with education on the new approach as well as indicating what not to use.

### DIFF
--- a/projects/go-lib/src/lib/components/go-modal/go-modal.component.spec.ts
+++ b/projects/go-lib/src/lib/components/go-modal/go-modal.component.spec.ts
@@ -69,7 +69,7 @@ describe('GoModalComponent', () => {
     });
 
     it('handles when no title and no modal size is set', () => {
-      component.currentComponent = new GoModalItem(GoTestModalHostComponent, {});
+      component.currentComponent = { component: GoTestModalHostComponent, bindings: { } };
 
       expect(component.modalTitle).toBeUndefined();
       expect(component.modalSize).toEqual(component.defaultModalSize);
@@ -81,7 +81,11 @@ describe('GoModalComponent', () => {
     });
 
     it('handles when a title is set', () => {
-      component.currentComponent = new GoModalItem(GoTestModalHostComponent, { modalTitle: 'Test Title' });
+      component.currentComponent = {
+        component: GoTestModalHostComponent,
+        bindings: { },
+        modalOptions: { modalTitle: 'Test Title' }
+      };
 
       expect(component.modalTitle).toBeUndefined();
 
@@ -91,7 +95,11 @@ describe('GoModalComponent', () => {
     });
 
     it('handles when a modal size is set to supported size', () => {
-      component.currentComponent = new GoModalItem(GoTestModalHostComponent, { modalSize: 'xl' });
+      component.currentComponent = {
+        component: GoTestModalHostComponent,
+        bindings: { },
+        modalOptions: { modalSize: 'xl' }
+      };
 
       expect(component.modalSize).toEqual(component.defaultModalSize);
 
@@ -100,18 +108,12 @@ describe('GoModalComponent', () => {
       expect(component.modalSize).toEqual('xl');
     });
 
-    it('handles when a modal size is set to unsupported size', () => {
-      component.currentComponent = new GoModalItem(GoTestModalHostComponent, { modalSize: 'abc' });
-
-      expect(component.modalSize).toEqual(component.defaultModalSize);
-
-      component.loadComponent();
-
-      expect(component.modalSize).toEqual(component.defaultModalSize);
-    });
-
     it('sets closeWithBackdrop to false if not passed in', () => {
-      component.currentComponent = new GoModalItem(GoTestModalHostComponent, {  });
+      component.currentComponent = {
+        component: GoTestModalHostComponent,
+        bindings: { }
+      };
+
       component.closeWithBackdrop = true;
 
       component.loadComponent();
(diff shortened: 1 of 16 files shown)


projects/go-lib/src/lib/components/go-modal/go-modal.component.spec.ts
```diff
@@ -100,18 +108,12 @@ describe('GoModalComponent', () => {
...
-    it('handles when a modal size is set to unsupported size', () => {
-      component.currentComponent = new GoModalItem(GoTestModalHostComponent, { modalSize: 'abc' });
-
-      expect(component.modalSize).toEqual(component.defaultModalSize);
-
-      component.loadComponent();
-
-      expect(component.modalSize).toEqual(component.defaultModalSize);
-    });
-
```
grahamhency, line 112: This test was removed because with the new type enforcements this can no longer be an issue.

projects/go-lib/src/lib/components/go-modal/go-modal.component.ts
```diff
@@ -11,24 +11,20 @@ import {
...
-  closeWithBackdrop: boolean = false;
-  currentComponent: any;
-  modalTitle: string;
-  modalSize: 'lg' | 'xl' = this.defaultModalSize;
-  noContentPadding: boolean = false;
+export class GoModalComponent extends GoModalOptions implements OnInit, OnDestroy {
+
+  currentComponent: GoModalItem<any>;
   opened: boolean = false;
-  showCloseIcon: boolean = true;
```
grahamhency, line 31: These are moved into the `GoModalOptions` class.

projects/go-lib/src/lib/components/go-modal/go-modal.component.ts
```diff
@@ -117,6 +93,37 @@ export class GoModalComponent implements OnInit, OnDestroy {
     return { 'go-modal--visible': this.opened };
   }
 
+  private setModalProperties(componentRef: ComponentRef<{}>): void {
+    if (this.currentComponent.modalOptions) {
```
grahamhency, line 97: For clarity, this `if` determines that if a `modalOptions` object exists on the component object that was passed in that those properties will be used to render the modal. It also means that it will ignore any properties in the `bindings` object that might pertain to the modal.

This is by design. We either use the old way or the new way, not both.

projects/go-lib/src/lib/components/go-modal/go-modal.service.ts
```diff
@@ -1,24 +1,35 @@
...
 
-import { GoModalItem } from './go-modal.item';
+import { GoModalBindings, GoModalItem } from './go-modal.item';
 import { Subject } from 'rxjs';
+import { GoModalOptions } from './go-modal-options';
 
 @Injectable()
 export class GoModalService {
-  activeModalComponent: Subject<GoModalItem> = new Subject<GoModalItem>();
+  activeModalComponent: Subject<GoModalItem<any>> = new Subject<GoModalItem<any>>();
```
grahamhency, line 9: Have to declare type `any` here until it is set in the service.

projects/go-style-guide/src/app/features/ui-kit/components/off-canvas-docs/off-canvas-docs.component.ts
```diff
@@ -58,9 +58,11 @@ export class OffCanvasDocsComponent {
     this.goOffCanvasService.openOffCanvas({
       component: BasicTestComponent,
       bindings: {
-        someBinding: 'monkey'
```
grahamhency, line 61: This is an example of how you could pass anything into the Components/Off-Canvas. This property didn't exist anywhere on the Off Canvas or BasicTestComponent.

StevenUlmer (changes requested):
This all looks great. The one thing I would say is we should probably keep the tests around for the deprecated behavior until it's actually removed.

projects/go-lib/src/lib/components/go-modal/go-modal.component.spec.ts
```diff
@@ -81,7 +81,11 @@ describe('GoModalComponent', () => {
     });
 
     it('handles when a title is set', () => {
-      component.currentComponent = new GoModalItem(GoTestModalHostComponent, { modalTitle: 'Test Title' });
```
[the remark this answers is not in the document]
grahamhency: That's a good idea, I went ahead and added tests for the deprecated functionality and noted it as such.

grahamhency:
@StevenUlmer ready for a re-review here.

StevenUlmer (approved):
Looks good to me

PR #733, opened 2021-01-25
[Bug] Table SelectAll Subscription

grahamhency:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

Whenever the `tableConfig` of the `go-table` is assigned (or reassigned) the table subscribes to the `selectAllControl`. This results in multiple subscriptions on the single control. We watch for changes to this control and emit an event whenever it changes. This is problematic because we will emit the event for each subscription. 

This is especially problematic when pagination or sorting happens.

## What is the new behavior?
This moves the subscription into the `ngOnInit` so that we only subscribe to the changes once per lifetime of the table.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/karma.conf.js
+++ b/projects/go-lib/karma.conf.js
@@ -16,9 +16,9 @@ module.exports = function (config) {
     reporters: ['dots'],
     port: 9876,
     colors: true,
-    logLevel: config.LOG_INFO,
+    logLevel: config.LOG_WARN,
     autoWatch: true,
-    browsers: ['Chrome'],
+    browsers: ['ChromeHeadless'],
     singleRun: false
   });
 };
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.spec.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.spec.ts
@@ -391,7 +391,7 @@ describe('GoTableComponent', () => {
 
       component.selectAllControl.setValue(false);
 
-      component['selectionSubscriptions'].unsubscribe();
+      component.ngOnDestroy();
 
       expect(component.selectAllIndeterminate).toBe(false);
     });
@@ -410,7 +410,7 @@ describe('GoTableComponent', () => {
 
       component.selectAllControl.setValue(true);
 
-      component['selectionSubscriptions'].unsubscribe();
+      component.ngOnDestroy();
 
       expect(component.targetedRows).toEqual([]);
     });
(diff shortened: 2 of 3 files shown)


projects/go-lib/karma.conf.js
```diff
@@ -16,9 +16,9 @@ module.exports = function (config) {
     reporters: ['dots'],
     port: 9876,
     colors: true,
-    logLevel: config.LOG_INFO,
+    logLevel: config.LOG_WARN,
     autoWatch: true,
-    browsers: ['Chrome'],
+    browsers: ['ChromeHeadless'],
```
grahamhency, line 21: I was using these to make testing a little easier, not really related to the PR.. but probably worth changing regardless.

projects/go-lib/src/lib/components/go-table/go-table.component.spec.ts
```diff
@@ -149,18 +149,6 @@ describe('GoTableComponent', () => {
...
-      component.selectAllControl.setValue(false);
-    });
-
-    it('should select all rows if preselected', () => {
-      component.tableConfig.preselected = true;
-      component.ngAfterViewInit();
-      expect(component.selectAllControl.value).toBe(true);
-    });
-  });
-
```
grahamhency, line 163: Again, not really related to this PR... but we have a block for `ngAftgerViewInit` at the bottom of the page and I am consolidating.

projects/go-lib/src/lib/components/go-table/go-table.component.spec.ts
```diff
@@ -399,8 +387,12 @@ describe('GoTableComponent', () => {
     });
 
     it('should set selectAllIndeterminate to false if selectAll toggled off', () => {
+      component.ngOnInit();
+
       component.selectAllControl.setValue(false);
 
+      component['selectionSubscriptions'].unsubscribe();
```
StevenUlmer, line 394: Does it not work if we call `component.ngOnDestroy()` here? I believe throughout goponents we're using `takeUntil` to clean up after subscriptions instead of using the unsubscribe method.

StevenUlmer (changes requested):
I think the logic here is fine, but I'm curious why we're adding the subscription to the component. We handle subscriptions the same way other places, so if it's not working we will likely need to make the same change in other places as well.

projects/go-lib/src/lib/components/go-table/go-table.component.spec.ts
grahamhency: Yeahhhh, you're right. This was like my 20th attempt to fix the issue so I forgot about cleaning this up lol. I switched everything to use the `ngOnDestroy` method instead and was able to remove the subscription.

StevenUlmer (approved):
Makes sense to me

PR #750, opened 2021-02-15
[Bug] Table False Values

### LINKED ISSUE #681: [BUG] GoTable doesnt display 'false' values for a boolean type column.
**Describe the bug**
Lets say there is a column in the go-table, which is of type boolean i.e. that particular field in the API is boolean, the go-table shows only the values that are `true`. It displays an empty string if the value is `false`. 

As a workaround, if you include an ng-template for that column, the `false` values start to show up.
But if in case you want to show `false` as `False` or `FALSE` (basically you want to pipe it), the table logic by-passes that expression/ logic.

**To Reproduce**
1. In your implementation of the go-table, include a column that displays boolean values.
2. Feed data into it with a mix of true and false values. Check the UI. You'll notice that only truthy (`true`) values are visible.

**Expected behavior**
The go-column should display `false` values.
Suggestion, we can replace `true/ false` with `Yes/No` to avoid any reserved keyword clashes, in case it makes things easier.

grahamhency:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

If a table column value was a `false` boolean it would be returned and displayed as `''`.
Resolves #681 

## What is the new behavior?
If table column is of type `boolean`, it returns the value as the string version of a boolean.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table-utils.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table-utils.ts
@@ -1,24 +1,14 @@
 export function extractFieldData(key: string, obj: object): any {
   if (key) {
-    return key.split('.').reduce((p: any, c: any) => {
-      let val: any = p && p[c];
-
-      if (typeof val === 'boolean') {
-        val = val.toString();
-      }
-
-      return val;
-    }, obj) || '';
+    return key.split('.').reduce((p: any, c: any) => p && p[c], obj);
   }
   return '';
 }
 
 export function sortBy(key: string, reverse: boolean): (a: object, b: object) => 1 | 0 | -1 {
   return (a: object, b: object) => {
-    let aFieldData: any = extractFieldData(key, a);
-    let bFieldData: any = extractFieldData(key, b);
-    aFieldData = typeof aFieldData === 'string' ? aFieldData.toLowerCase() : aFieldData;
-    bFieldData = typeof bFieldData === 'string' ? bFieldData.toLowerCase() : bFieldData;
+    const aFieldData: any = formatSortValue(key, a);
+    const bFieldData: any = formatSortValue(key, b);
 
     if (aFieldData < bFieldData) {
       return reverse ? -1 : 1;
@@ -30,4 +20,14 @@ export function sortBy(key: string, reverse: boolean): (a: object, b: object) =>
 
     return 0;
   };
+}
+
+export function formatSortValue(key: any, object: any): any {
+  const value: any = nullCheck(extractFieldData(key, object));
+  return typeof value === 'string' ? value.toLowerCase() : value;
+}
+
+// TODO: Replace instances using this with ?? after upgrade to typescript 3.7+ and tsconfig.target is ES2020
+export function nullCheck(value: any, fallback: any = ''): any {
+  return value !== null && value !== undefined ? value : fallback;
 }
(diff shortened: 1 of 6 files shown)


projects/go-lib/src/lib/components/go-table/go-table-utils.ts
```diff
@@ -1,11 +1,19 @@
...
+      }
+
+      return val;
+    }, obj) || '';
   }
   return '';
 }
 
-export function sortBy(key: string, reverse: boolean): (a: object, b: object) => number {
+export function sortBy(key: string, reverse: boolean): (a: object, b: object) => 1 | 0 | -1 {
```
patrick-lewandowski, line 16: 👍

projects/go-lib/src/lib/components/go-table/go-table-utils.ts
```diff
@@ -1,11 +1,19 @@
...
-    return key.split('.').reduce((p: any, c: any) => p && p[c], obj) || '';
+    return key.split('.').reduce((p: any, c: any) => {
+      let val: any = p && p[c];
+
+      if (typeof val === 'boolean') {
+        val = val.toString();
+      }
+
+      return val;
+    }, obj) || '';
```
patrick-lewandowski, line 11: This logic still prevents a numeric `0` from showing since `0 || ''` will resolve to `''`. 

You can test this in the docs by removing the `+1` from https://github.com/mobi/goponents/blob/bug_table_false_values/projects/go-style-guide/src/app/features/ui-kit/components/table-docs/table-docs.service.ts#L21 so a `0` id will generate. 

Is this meant to also be fixed by this PR or will it continue to be an open issue?

grahamhency: @patrick-lewandowski I thought there was something else, but I couldn't remember what it was and we didn't have an actual issue for it. I can look into that..

Honestly, maybe we should just `.toString` everything? If you need extra formatting that's what the templates are for anyway.

patrick-lewandowski: The other option is take out the `|| ''` part and have everything return as expected. If it's meant to be `null` or `undefined`, then so be it. I personally don't think we should be assuming everything should default to an empty string. You can try `toString` just not sure if there would be any edge cases around it.

grahamhency: That's a good point. The issue we ran into in the past when we tried to remove the `|| ''` was that it broke client side sorting because any of the columns that contained `null` or `undefined` values weren't able to be sorted.

Taking that into account, I removed the `|| ''` from this method and updated the `sortBy` method to account for `null` or `undefined` values and default them to `''` strictly inside of the `sortBy` method.

This way we're not modifying any data, correcting both bugs, and satisfying all the cases (I think).

patrick-lewandowski (approved):
LGTM - in the release notes I would mention that `0` works correctly now and instances of `'0'` should be addressed to avoid any sorting issues since a number zero vs string zero might not behave as expected (didn't test it out myself). I know in command center for example they convert zeros to string to get around this issue in some cases, not sure if other apps did the same.

grahamhency:
@patrick-lewandowski that is a good idea to mention, thanks for the cross platform knowledge there. I wish the issue would've been resolved in Goponents instead of patching it in Command Center.. but oh well. It'll be fixed now.

PR #754, opened 2021-02-16
[Feature] Button Groups

### LINKED ISSUE #688: [Feature] Button Groups
We would like to add button groups to the design system, specifically ones that connect.

Please see documentation here: https://tangoe.invisionapp.com/dsm/tangoe/tangoe-dsm/nav/build/folder/components/5f91b212c950fa0597809ffb

grahamhency:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

Buttons cannot be grouped in the manner UX is specifying
Resolves #688 

## What is the new behavior?
A new component that allows you to specify a group of buttons that share variants and disabled states.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-button-group/go-button-group.component.spec.ts
+++ b/projects/go-lib/src/lib/components/go-button-group/go-button-group.component.spec.ts
@@ -1,0 +1,156 @@
+import { ElementRef, QueryList } from '@angular/core';
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { GoButtonComponent } from '../go-button/go-button.component';
+import { GoButtonGroupComponent } from './go-button-group.component';
+
+describe('GoButtonGroupComponent', () => {
+  let component: GoButtonGroupComponent;
+  let fixture: ComponentFixture<GoButtonGroupComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ GoButtonGroupComponent ]
+    })
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(GoButtonGroupComponent);
+    component = fixture.componentInstance;
+    component.buttons = new QueryList();
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  describe('ngOnChanges', () => {
+    it('calls processButtons', () => {
+      spyOn<any>(component, 'processButtons').and.callThrough();
+
+      component.ngOnChanges();
+
+      expect(component['processButtons']).toHaveBeenCalled();
+    });
+  });
+
+  describe('ngAfterContentInit', () => {
+    it('calls processButtons', () => {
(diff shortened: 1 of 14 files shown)


projects/go-lib/src/lib/components/go-button-group/go-button-group.component.spec.ts
```diff
@@ -0,0 +1,157 @@
+import { ElementRef, QueryList } from '@angular/core';
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { GoButtonComponent } from '../go-button/go-button.component';
+
```
hoosierplew, line 4: Only asking as this is my first PR in this repo - do we typically have blank lines between some imports?

projects/go-lib/src/lib/components/go-button-group/go-button-group.component.ts
```diff
@@ -0,0 +1,54 @@
...
+        btn.groupPosition = this.determinePosition(i);
+        btn.buttonDisabled = this.buttonDisabled;
+        btn.buttonVariant = this.buttonVariant;
+        btn.ngOnChanges();
+      });
+    }
+  }
+
+  private determinePosition(i: number): GoButtonComponent['groupPosition'] {
+    let position: GoButtonComponent['groupPosition'] = null;
```
hoosierplew, line 35: If `determinePosition` is expected to always return either `first`, `last`, or `middle`, you could rewrite this function without using the `position` variable and reduce complexity greatly:

```Javascript
if (this.buttons.length > 1) {
  if (i === 0) {
    return 'first';
  }
  
  if (i === this.buttons.length -1) {
    return 'last';
  }
  
  return 'middle';
}
```

Not sure if this is proper formatting for this codebase, but it _might_ remove the cognitive complexity warning.

hoosierplew (dismissed):
Approving as is, but it would be nice if we could get that complexity score down for `determinePosition()`.

projects/go-lib/src/lib/components/go-button-group/go-button-group.component.spec.ts
grahamhency: @hoosierplew sometimes we do (we do this in `mobi` as well). If there are a lot of imports in a file we will try to logically group them by third party vs. our own and/or by type of import (services, components, etc)

In this case, we don't need to do this. I'll remove the line.

projects/go-lib/src/lib/components/go-button-group/go-button-group.component.ts
grahamhency: Technically, it can also return `null`. If there is only 1 `go-button` in the `go-button-group` then we will want to return `null` so that it doesn't apply any of the modifier classes.

hoosierplew: Good catch, I added a `return null` to the end.

grahamhency: I updated it to what you put, I do think that's a little bit more concise.

PR #765, opened 2021-02-26
Add 'Dismissable' argument to Toaster Service

deotte:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

Currently, toasts that come in from the `goToasterService` aren't dimissable. The `GoToastComponent` supports a `dismissable` input, so we should add this to the service too.

## What is the new behavior?

The `goToasterService` supports an additional argument, `dismissable` through the `toastInterface`.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-toaster/go-toaster.service.ts
+++ b/projects/go-lib/src/lib/components/go-toaster/go-toaster.service.ts
@@ -75,6 +75,7 @@ export class GoToasterService {
     comp.message = toastInterface.message;
     comp.type = toastInterface.type;
     comp.header = toastInterface.header;
+    comp.dismissable = toastInterface.dismissable;
     comp.duration = duration;
 
     return  comp;
--- a/projects/go-style-guide/src/app/features/ui-kit/components/toast-docs/toast-docs.component.ts
+++ b/projects/go-style-guide/src/app/features/ui-kit/components/toast-docs/toast-docs.component.ts
@@ -88,6 +88,9 @@ export class ToastDocsComponent {
 
   // you can adjust the duration of the toast by passing in the duration param:
   this.toasterService.toastSuccess({ message: 'You clicked the button!' }, 7000);
+
+  // you can also make the toast dismissable by setting the "dismissable" property on the ToastInterface param:
+  this.toasterService.toastSuccess({ message: 'You clicked the button!', dimissable: true });
   `;
 
   toaster_app_ts: string = `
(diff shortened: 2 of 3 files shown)


projects/go-lib/src/lib/components/go-toaster/go-toaster.service.ts
```diff
@@ -75,6 +75,7 @@ export class GoToasterService {
     comp.message = toastInterface.message;
     comp.type = toastInterface.type;
     comp.header = toastInterface.header;
+    comp.dismissable = toastInterface.dismissable !== undefined ? toastInterface.dismissable : false;
```
grahamhency, line 78: We can actually simplify this in JavaScript land:
```ts
comp.dismissable = toastInterface.dismissable;
```
If `dismissable` is `undefined`, `null`, or `false` it will evaluate to `false` and because it's strictly typed to be a `boolean` value it should only ever be set to `true` or default to `false` (whether it's actually `false` or just `null/undefined`).

projects/go-style-guide/src/app/features/ui-kit/components/toast-docs/toast-docs.component.ts
```diff
@@ -88,6 +88,9 @@ export class ToastDocsComponent {
 
   // you can adjust the duration of the toast by passing in the duration param:
   this.toasterService.toastSuccess({ message: 'You clicked the button!' }, 7000);
+
+  // you can also make the toast dismissable by passing in a "dismissable" param to toastSuccess:
```
grahamhency, line 92: I assume this doesn't just work for `toastSuccess`, but anything where we're using the `ToastInterface` right? Should we make this documentation a little more general then?
```ts
// you can also make the toast dismissable by setting the "dismissable" property on the ToastInterface param

grahamhency (changes requested):

projects/go-lib/src/lib/components/go-toaster/go-toaster.service.ts
deotte: Oh that's awesome! Thanks

projects/go-style-guide/src/app/features/ui-kit/components/toast-docs/toast-docs.component.ts
deotte: Yep! I wanted it to put it in the interface so that it could be used anywhere. That looks good though, I'll implement that 👍

grahamhency (approved):
Thanks for contributing!

PR #773, opened 2021-03-09
[Bug] Go-Select Multi-Select errors when "items" null

grahamhency:
## PR Checklist

~~Tests for the changes have been added~~

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

In the `go-select`, when `multiple` is `true` and `items` are `null` it is throwing a console error saying `cannot read property "length" of undefined".

## What is the new behavior?
It no longer does that

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-select/go-select.component.html
+++ b/projects/go-lib/src/lib/components/go-select/go-select.component.html
@@ -30,7 +30,7 @@
   [virtualScroll]="virtualScroll"
   (scrollToEnd)="onScrollToEnd()"
   (scroll)="onScroll($event)">
-  <ng-container *ngIf="multiple && showSelectAll && items?.length > 0">
+  <ng-container *ngIf="multiple && showSelectAll && items?.length">
     <ng-template ng-header-tmp>
       <button
         (click)="onSelectAll()"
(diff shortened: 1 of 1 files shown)


projects/go-lib/src/lib/components/go-select/go-select.component.html
```diff
@@ -30,7 +30,7 @@
   [virtualScroll]="virtualScroll"
   (scrollToEnd)="onScrollToEnd()"
   (scroll)="onScroll($event)">
-  <ng-container *ngIf="multiple && showSelectAll && items.length > 0">
+  <ng-container *ngIf="multiple && showSelectAll && items?.length > 0">
```
hoosierplew, line 33: This works but now that we're qualifying that `items` is non-null you don't need to have the `> 0` check. `items?.length` should be enough.

hoosierplew (dismissed):
Good as it is, but could have one small tweak if you wanted to do it.

projects/go-lib/src/lib/components/go-select/go-select.component.html
grahamhency: Good point, I removed it.

PR #776, opened 2021-03-15
[Bug] Go-Table Change Selectable Issue

### LINKED ISSUE #774: [BUG] Table Change Selection Behavior
**Describe the bug**
On the `go-table`, when the table renders with `selectable: false` on the `tableConfig` the table renders fine. However, after it renders fine, if the `tableConfig` is reset with `selectable: true` then the selection behavior / events of the table are completely broken. 

**To Reproduce**
https://stackblitz.com/edit/goponents-table-change-selection-issue

**Expected behavior**
We should be able to switch the table between a selectable and not selectable state and have it work as intended regardless.

**Screenshots**
N/A

grahamhency:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

See the issue referenced below for description of problem.
Resolves #774 

## What is the new behavior?
The table will handle when it becomes `selectable` after not being `selectable` and vice versa.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table.component.ts
@@ -19,7 +19,7 @@ import {
   FormControl,
   FormGroup
 } from '@angular/forms';
-import { Subject, Subscription } from 'rxjs';
+import { Subject } from 'rxjs';
 import { debounceTime, distinctUntilChanged, takeUntil } from 'rxjs/operators';
 
 import { GoConfigService } from '../../go-config.service';
@@ -107,6 +107,7 @@ export class GoTableComponent implements OnInit, OnChanges, OnDestroy, AfterView
 
   private destroy$: Subject<void> = new Subject();
   private pageChange$: Subject<void> = new Subject();
+  private selections$: Subject<void> = new Subject();
 
   constructor(
     private changeDetector: ChangeDetectorRef,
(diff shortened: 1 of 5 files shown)


projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -107,6 +107,7 @@ export class GoTableComponent implements OnInit, OnChanges, OnDestroy, AfterView
 
   private destroy$: Subject<void> = new Subject();
   private pageChange$: Subject<void> = new Subject();
+  private selectionSubscriptions: Subscription = new Subscription();
```
StevenUlmer, line 110: I think everywhere else we've been adding a subject like `pageChange$` to track and remove subscriptions. Should we use the same here?

StevenUlmer, line 110: Do we need to clean up this subscription in the `ngOnDestroy` method?

StevenUlmer (changes requested):
Really the only concern here is around cleaning up the subscription on destroy. Could we also get a test around this change?

grahamhency:
@StevenUlmer I changed the approach up a bit to the `Subject` approach. I also added tests for this. Ready for another review.

StevenUlmer (approved):
Looks good. Thanks for adding the tests

PR #771, opened 2021-03-08
Integrate toast functionality for dark theme

amitpatra91:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our [guidelines](https://github.com/mobi/goponents/blob/master/CONTRIBUTING.md)
- [ ] Tests for the changes have been added
- [x] Docs have been added or updated

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation content changes
- [x] Style Update (CSS)
- [ ] Other... Please describe:

## What is the current behavior?
**go-toast**  has no ux features for dark theme

Resolves [737](https://github.com/mobi/goponents/issues/737)

## What is the new behavior?
The ability to style a **go-toast** for a dark theme.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #769, opened 2021-03-04
[Bug] Datepicker Not Resetting Correctly

SapanaTangoe:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our [guidelines](https://github.com/mobi/goponents/blob/master/CONTRIBUTING.md)
- [x] Tests for the changes have been added
- [x] Docs have been added or updated

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Style Update (CSS)
- [ ] Other... Please describe:

## What is the current behavior?

We are not able to reset the form such date picker control.

Resolves : GSCMD-736

## What is the new behavior?
We are able reset the form such date picker control. Because we have implemented new reset functionality.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #779, opened 2021-03-22
v1.10.0

grahamhency:
###### Change log for `v1.10.0`

***

# New Features

### Toaster
- Added ability to pass `dismissable` to the toasts in the toaster #765 
- Added support for a dark theme #771 

# Bug Fixes

### Datepicker
- Fixed an issue where resetting the `FormControl` would not properly reset the datepicker #769 

### Select
- Fixed an issue where the select would error when `items` were `null` and `multiple` was `true` #773 

### Table
- Fixed an issue where the table would error if `selectable` changed when the `tableConfig` is reset #776


patrick-lewandowski (approved):
LGTM

PR #780, opened 2021-03-23
Support Sticky Headers 777

### LINKED ISSUE #777: [Feature] Support Sticky Headers in go-table
As a user, I would like the ability to pass an option to `go-table` to make all column headers sticky. 

_**Example:**_
An example can be found here: https://material.angular.io/components/table/examples#table-sticky-complex
However, the ask here is just for the **sticky headers** support.

I would like to pass a `boolean` to some option in `go-table`, for example `stickyHeaders`. When `true`, all column headers should remain sticky when the data is scrollable. 

An example of when this would be necessary is when `maxHeight` is defined and the number of rows causes the table height to be greater than the max height defined. If a user scrolls, they lose sight of column headers and it becomes hard to tell the purpose of each column in the row. With this new option, the headers should scroll with the data like in the example link above.

SapanaTangoe:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Style Update (CSS)

when scrolled table contents then table header are hide in component

Resolves #777 

## What is the new behavior?
In this change table header position sticky on top. Meaning, when scrolled, the sticky headers will stay where it is.

- [ ] Yes
- [x] No


projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -38,14 +38,14 @@ <h3 class="go-heading-3 go-heading--no-margin go-table-header__title" *ngIf="tab
...
-            class="go-table__body-column"
+          class="go-table__body-column"
             *ngIf="details || childRows"
             width="40">
           </th>
 
           <!-- Header - Select All Checkbox Column -->
           <th
-            class="go-table__head-column go-table__head-checkbox go-table__head--sortable"
+            class="go-table__head-column go-table__head-checkbox go-table__head--sortable {{classStickyHeader}}"
```
grahamhency, line 48: There really should never be a reason to interpolate things into the `class` attribute. For this you'll want to use `ngClass`.  See [the docs](https://angular.io/api/common/NgClass).

projects/go-lib/src/lib/components/go-table/go-table.component.html
```diff
@@ -57,7 +57,7 @@ <h3 class="go-heading-3 go-heading--no-margin go-table-header__title" *ngIf="tab
 
           <!-- Header - Columns -->
           <th *ngFor="let col of columns"
-              class="go-table__head-column"
+              class="go-table__head-column {{classStickyHeader}}"
```
grahamhency, line 60: Same here, use `ngClass` instead.

projects/go-lib/src/lib/components/go-table/go-table.component.spec.ts
```diff
@@ -1190,4 +1206,5 @@ describe('GoTableComponent', () => {
       expect(component['changeDetector'].detectChanges).toHaveBeenCalled();
     });
   });
+
```
grahamhency, line 1209: Unnecessary change.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -104,6 +104,10 @@ export class GoTableComponent implements OnInit, OnChanges, OnDestroy, AfterView
   selectAllIndeterminate: boolean = false;
   showTable: boolean = false;
   targetedRows: any[] = [];
+  stickyHeader: boolean = false;
+  classStickyHeader: string[];
+  stickyClass: string = 'go-table__head-column-sticky';
+  nonStickyClass: string = 'go-table__head-column-nonsticky';
```
grahamhency, line 110: We do not need these variables.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -592,6 +597,11 @@ export class GoTableComponent implements OnInit, OnChanges, OnDestroy, AfterView
     });
   }
 
+  private setupSticky(): void {
+      this.stickyHeader = Boolean(this.tableConfig.stickyHeader);
+      this.classStickyHeader = this.stickyHeader ? [this.stickyClass] : [this.nonStickyClass];
+    }
+
```
grahamhency, line 604: This can all be deleted.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -173,7 +178,7 @@ export class GoTableComponent implements OnInit, OnChanges, OnDestroy, AfterView
     }
 
     this.showTable = Boolean(this.tableConfig);
-  }
+    }
```
grahamhency, line 181: Unnecessary change.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -126,10 +130,12 @@ export class GoTableComponent implements OnInit, OnChanges, OnDestroy, AfterView
       this.setupPageSizes();
       this.setupConfigService();
       this.setupSelectAllControlSub();
+     this.setupSticky();
```
grahamhency, line 133: This can be deleted.

projects/go-lib/src/lib/components/go-table/go-table.component.ts
```diff
@@ -126,10 +130,12 @@ export class GoTableComponent implements OnInit, OnChanges, OnDestroy, AfterView
...
     }
   }
 
   ngOnChanges(changes: SimpleChanges): void {
+    this.setupSticky();
```
grahamhency, line 138: This can be deleted.

projects/go-lib/src/lib/components/go-table/go-table.component.spec.ts
```diff
@@ -1119,6 +1119,14 @@ describe('GoTableComponent', () => {
...
       expect(component.tableChange.emit).toHaveBeenCalled();
     }));
+
+    it('set sticky header on table', fakeAsync(() => {
+      component.tableConfig.dataMode = GoTableDataSource.server;
+      component.tableConfig.stickyHeader = true;
+      component.ngOnInit();
+      expect(component.tableConfig.stickyHeader).toEqual(component.stickyHeader);
+    }));
+
```
grahamhency, line 1129: If we adopt the `@Input` approach then these tests are no longer necessary.

projects/go-lib/src/lib/components/go-table/go-table.component.spec.ts
```diff
@@ -1129,6 +1137,14 @@ describe('GoTableComponent', () => {
...
       expect(component.renderTable).toHaveBeenCalled();
     });
+
+    it('set sticky header on table', fakeAsync(() => {
+      component.tableConfig.dataMode = GoTableDataSource.server;
+      component.tableConfig.stickyHeader = false;
+      component.ngOnChanges({ tableConfig: {} as SimpleChange });
+      expect(component.tableConfig.stickyHeader).toEqual(component.stickyHeader);
+    }));
+
```
grahamhency, line 1147: If we adopt the `@Input` approach then these tests are no longer necessary.

projects/go-style-guide/src/app/features/ui-kit/components/table-docs/table-docs.component.ts
```diff
@@ -27,7 +27,8 @@ export class TableDocsComponent {
       { route: './searching', routeTitle: 'Searching' },
       { route: './server-integration', routeTitle: 'Server-Side Integration' },
       { route: './sorting', routeTitle: 'Sorting' },
-      { route: './title-template', routeTitle: 'Title Template' }
+      { route: './title-template', routeTitle: 'Title Template' },
+      { route: './sticky', routeTitle: 'Sticky Table' }
```
grahamhency, line 31: Would you mind alphabetizing these by the `routeTitle`?

grahamhency (changes requested):
I am requesting changes because the logic proposed in this PR vastly overcomplicates the functionality required for the feature. I've proposed an alternatively which would allow us to remove the unit tests and additional logic in the `go-table.component.ts` file as well.

I would suggest making the necessary changes to this PR to reflect that alternative approach.

You will need to change all of the documentation to reflect this. Because we're just adding a binding to the `GoTableComponent` all that needs to happen for the documentation is to add the appropriate documentation to the base `table-overview.component` section. We won't need a new route for this.

projects/go-lib/src/lib/components/go-table/go-table.component.scss
```diff
@@ -74,6 +73,13 @@ $go-table-body-row-expanded-border: rgba(240, 240, 240, 0);
     &:last-of-type {
       border-top-right-radius: $global-radius;
     }
+
+    &--sticky {
+      @include z-index(2);
+      background: $theme-light-bg;
+      position: sticky;
```
grahamhency, line 80: _technically_ this will not work in IE. I think we're okay with still using this though. Because IE is being deprecated, in this case the fallback is that the header will behave as it does in the default.

grahamhency (changes requested):
Getting closer here, but still some minor fixes and edge cases you'll need to address. The top border in the table content needs to be moved to the `th` elements. As we scroll the border moves with the table content. It should stick to the headers. See the recording below:

https://user-images.githubusercontent.com/11966331/114184238-9993e680-9912-11eb-8ce1-2199c6876c58.mov

projects/go-lib/src/lib/components/go-table/go-table.component.scss
```diff
@@ -74,6 +74,14 @@ $go-table-body-row-expanded-border: rgba(240, 240, 240, 0);
     &:last-of-type {
       border-top-right-radius: $global-radius;
     }
+
+    &--sticky {
+      @include z-index(2);
+      background: $theme-light-bg;
+      box-shadow: inset 0 -2px 0 $theme-light-app-bg;
```
grahamhency, line 81: This doesn't need added to the `--sticky` class since we're replacing it in the base `__head-column` class.

projects/go-lib/src/lib/components/go-table/go-table.component.scss
```diff
@@ -74,6 +73,13 @@ $go-table-body-row-expanded-border: rgba(240, 240, 240, 0);
     &:last-of-type {
       border-top-right-radius: $global-radius;
     }
+
+    &--sticky {
+      @include z-index(2);
```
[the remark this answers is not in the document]
grahamhency: I don't see any changes that would solve this. Can you check this?

grahamhency (changes requested):

projects/go-lib/src/lib/components/go-table/go-table.component.scss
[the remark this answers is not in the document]
grahamhency: This was resolved by adding extra `z-index` to the `go-table-loader` class.

grahamhency (approved):
Looks good. :+1:

PR #816, opened 2021-08-04
added remove all on select

priyankafnu:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

When a lot of results are selected, the "clear all" button appears way down the page in an unusual position.

## What is the new behavior?
This adds a "remove all" button at the top of the select dropdown when items are selected to allow a user to easily remove all options when lots are selected.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-select/go-select.component.html
+++ b/projects/go-lib/src/lib/components/go-select/go-select.component.html
@@ -41,7 +41,7 @@
       </button>
     </ng-template>
   </ng-container>
-  <ng-container *ngIf="multiple && showRemoveAll && items?.length">
+  <ng-container *ngIf="multiple && items?.length">
     <ng-template ng-header-tmp>
       <button
       (click)="onRemoveAll()"
--- a/projects/go-lib/src/lib/components/go-select/go-select.component.html
+++ b/projects/go-lib/src/lib/components/go-select/go-select.component.html
@@ -39,6 +39,10 @@
         [ngClass]="{ 'go-select__select-all-button--dark' : theme === 'dark' }">
         Select All
       </button>
+    </ng-template>
+  </ng-container>
+  <ng-container *ngIf="multiple && showRemoveAll && items?.length">
+    <ng-template ng-header-tmp>
       <button
       (click)="onRemoveAll()"
       class="go-select__select-all-button"
--- a/projects/go-lib/src/lib/components/go-select/go-select.component.html
+++ b/projects/go-lib/src/lib/components/go-select/go-select.component.html
@@ -39,6 +39,12 @@
         [ngClass]="{ 'go-select__select-all-button--dark' : theme === 'dark' }">
         Select All
       </button>
+      <button
+      (click)="onRemoveAll()"
+      class="go-select__select-all-button"
+      [ngClass]="{ 'go-select__select-all-button--dark' : theme === 'dark' }">
+      Remove  All
+    </button>
     </ng-template>
   </ng-container>
   <ng-container *ngIf="goSelectSelectedOption">
--- a/projects/go-lib/src/lib/components/go-select/go-select.component.html
+++ b/projects/go-lib/src/lib/components/go-select/go-select.component.html
@@ -41,7 +41,7 @@
       </button>
     </ng-template>
   </ng-container>
-  <ng-container *ngIf="multiple && items?.length">
+  <ng-container *ngIf="multiple && items?.length && control.value">
     <ng-template ng-header-tmp>
       <button
       (click)="onRemoveAll()"
--- a/projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts
+++ b/projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts
@@ -104,13 +104,15 @@ describe('GoSelectComponent', () => {
   });
 
   describe('onRemoveAll', () => {
-    it('uses bindValue to get value if bindValue exists', () => {
+    it('uses removed the selected values', () => {
       component.bindValue = 'id';
       component.items = [
         { id: 1, label: 'Label 1' },
         { id: 2, label: 'Label 2' },
         { id: 3, label: 'Label 3' }
       ];
+
+      component.onSelectAll();
 
       component.onRemoveAll();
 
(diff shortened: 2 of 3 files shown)


projects/go-lib/src/lib/components/go-select/go-select.component.html
```diff
@@ -39,6 +39,12 @@
         [ngClass]="{ 'go-select__select-all-button--dark' : theme === 'dark' }">
         Select All
       </button>
+      <button
+      (click)="onRemoveAll()"
+      class="go-select__select-all-button"
+      [ngClass]="{ 'go-select__select-all-button--dark' : theme === 'dark' }">
+      Remove  All
+    </button>
```
grahamhency, line 47: Should we have a conditional here to only render the button if the value of the control is set? Not sure it makes sense from a UX perspective to show a "Remove All" option if they haven't selected anything.

projects/go-lib/src/lib/components/go-select/go-select.component.ts
```diff
@@ -56,6 +56,10 @@ export class GoSelectComponent implements OnInit {
     this.control.patchValue(this.items.map((item: any) => this.bindValue ? item[this.bindValue] : item));
   }
 
+  onRemoveAll(): void {
+    this.control.reset();
+  }
```
grahamhency, line 61: Should we write a test for this method?

grahamhency (changes requested):
Functionality looks good. I had one additional suggestion and a comment on tests.

projects/go-lib/src/lib/components/go-select/go-select.component.html
priyankafnu: @grahamhency I have added showRemoveAll condition. Please check and let me know if it is okay or not

projects/go-lib/src/lib/components/go-select/go-select.component.ts
priyankafnu: @grahamhency test case done. Please check

projects/go-lib/src/lib/components/go-select/go-select.component.html
grahamhency: @priyankafnu my question still stands. The change that you've put up to add the `showRemoveAll` binding here will always be true unless someone passes in the value to `go-select` of `true`. I don't think we need the `showRemoveAll` binding here. We can probably just do:
```
<ng-container *ngIf="multiple && items?.length">

File: projects/go-lib/src/lib/components/go-select/go-select.component.ts [hunk 3]
```diff
@@ -35,6 +35,7 @@ export class GoSelectComponent implements OnInit {
   @Input() placeholder: string;
   @Input() searchable: boolean = true;
   @Input() showSelectAll: boolean = true;
+  @Input() showRemoveAll: boolean = false;
```
grahamhency, line 38: As stated in my other comment, I don't believe we need this. If you wanted to keep it to be able to override showing/hiding the button, then this should default to `true` and we can keep the binding in the conditional.

projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts
```diff
@@ -102,4 +102,19 @@ describe('GoSelectComponent', () => {
       expect(component.control.value).toEqual([1, 2, 3]);
     });
   });
+
+  describe('onRemoveAll()', () => {
```
grahamhency, line 106: This should just be `onRemoveAll`, the parenthesis are not needed.

projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts
```diff
@@ -102,4 +102,19 @@ describe('GoSelectComponent', () => {
...
+    it('uses bindValue to get value if bindValue exists', () => {
```
grahamhency, line 107: We should change the description here to be something related to what we're testing.

projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts
```diff
@@ -102,4 +102,19 @@ describe('GoSelectComponent', () => {
...
+      component.bindValue = 'id';
+      component.items = [
+        { id: 1, label: 'Label 1' },
+        { id: 2, label: 'Label 2' },
+        { id: 3, label: 'Label 3' }
+      ];
+
+      component.onRemoveAll();
```
grahamhency, line 115: Should we set `component.control` to a value prior calling this method to ensure that the control actually goes from a value to null?

grahamhency (changes requested):

projects/go-lib/src/lib/components/go-select/go-select.component.html
grahamhency: Also, we really should only show this button if the control has a value. So the final conditional should look like:
```
<ng-container *ngIf="multiple && items?.length && control.value">

File: projects/go-lib/src/lib/components/go-select/go-select.component.ts (hunk 3 above)
Comment (priyankafnu) [reply projects/go-lib/src/lib/components/go-select/go-select.component.ts]: Hi @grahamhency Done with the changes. Please check

File: projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts (hunk 4 above)
Comment (priyankafnu) [reply projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts]: Hi @grahamhency 
Removed parenthesis from onRemoveAll. Please check

File: projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts (hunk 6 above)
Comment (priyankafnu) [reply projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts:117]: Done this changes

File: projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts (hunk 5 above)
Comment (priyankafnu) [reply projects/go-lib/src/lib/components/go-select/go-select.component.spec.ts]: Done the changes

File: projects/go-lib/src/lib/components/go-select/go-select.component.html (hunk 1 above)
Comment (grahamhency) [reply projects/go-lib/src/lib/components/go-select/go-select.component.html:51]: @priyankafnu, I still think that we should only render this button when the control has a value. Would you mind adding it to the conditional?

```
<ng-container *ngIf="multiple && items?.length && control.value">

grahamhency (changes requested):
Just one more comment, but if you can address that I'll approve and we can get a new version ready.

grahamhency (approved):

PR #820, opened 2021-08-12
[Feature] go-table wrapContent

### LINKED ISSUE #811: [Feature] Go Table Column Wrap
I am requesting we add a binding on the `go-table-column` component to specify whether or not the column content will wrap or not.

On `go-table-column` add a binding such as `[wrapContent]` that takes a boolean and if true makes the content of that column wrap.

1. Add binding to `go-table-column` called `wrapContent` that is a `boolean`
2. `wrapContent` should default to `true`
3. If `wrapContent` is `false` it should apply the style `white-space: nowrap;` to the column's content

grahamhency:
## PR Checklist

~~- [ ] Tests for the changes have been added~~

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Style Update (CSS)

Resolves #811 

## What is the new behavior?
Added a binding to the `go-table` component called `wrapContent` that controls whether the content in all of the table cells wraps or not.

Added a binding to the `go-table-column` component called `wrapContent` that controls the content wrapping for that individual column. Note: this binding will override the `wrapContent` binding on the `go-table` for that individual column.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-table/go-table-column.component.ts
+++ b/projects/go-lib/src/lib/components/go-table/go-table-column.component.ts
@@ -11,6 +11,7 @@ export class GoTableColumnComponent extends GoTableChildColumnComponent {
   @Input() sortable?: boolean;
   @Input() title: string;
   @Input() width: number;
+  @Input() wrapContent: boolean = null;
 
   @ContentChild('goTableHead') goTableHead: TemplateRef<any>;
 
--- a/projects/go-style-guide/src/app/features/ui-kit/components/table-docs/components/table-column-docs/table-column-docs.component.ts
+++ b/projects/go-style-guide/src/app/features/ui-kit/components/table-docs/components/table-column-docs/table-column-docs.component.ts
@@ -20,6 +20,7 @@ export class TableColumnDocsComponent {
   @Input() sortable?:  boolean = true;
   @Input() title:      string;
   @Input() width:      number;
+  @Input() wrapContent: boolean = null;
   `;
 
   fieldExample: string = `
@@ -45,6 +46,9 @@ export class TableColumnDocsComponent {
     <go-table-column field="email" width="50"></go-table-column>
     <go-table-column field="gender"></go-table-column>
     <go-table-column field="ip_address" title="IP Address"></go-table-column>
+    <go-table-column field="last_login" title="Last Login" [wrapContent]="false"></go-table-column>
+    <go-table-column field="alive"></go-table-column>
+    <go-table-column field="login_attempts" title="Login Attempts"></go-table-column>
   </go-table>
   `;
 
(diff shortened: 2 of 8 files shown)


projects/go-lib/src/lib/components/go-table/go-table-column.component.ts
```diff
@@ -11,6 +11,7 @@ export class GoTableColumnComponent extends GoTableChildColumnComponent {
   @Input() sortable?: boolean;
   @Input() title: string;
   @Input() width: number;
+  @Input() wrapContent: boolean = null;
```
schneiderduketangoe, line 14: Would it be better to give this a default value of false? I'm not sure of the implications of doing it null v. false here but I feel like if the user didn't define the input then it would at least have a false fall-back value.

projects/go-style-guide/src/app/features/ui-kit/components/table-docs/components/table-column-docs/table-column-docs.component.ts
```diff
@@ -20,6 +20,7 @@ export class TableColumnDocsComponent {
   @Input() sortable?:  boolean = true;
   @Input() title:      string;
   @Input() width:      number;
+  @Input() wrapContent: boolean = null;
```
schneiderduketangoe, line 23: Same comment as the other file:

Would it be better to give this a default value of false? I'm not sure of the implications of doing it null v. false here but I feel like if the user didn't define the input then it would at least have a false fall-back value.

schneiderduketangoe (dismissed):
Looks good Graham. Small nit-picky comment but I'm not gonna let that hold back the approval

projects/go-lib/src/lib/components/go-table/go-table-column.component.ts
grahamhency: Good question. I ended up going the `null` route here because I wanted the ability to tell if the implementor specifically set the attribute on the column and because the wrapping could go both ways.

For Ex:
```html
<go-table [wrapContent]="true">
  <go-table-column field="id"></go-table-column>
  <go-table-column field="name" [wrapContent]="false"></go-table-column>
</go-table>
```
The `id` column will be null so it will use whatever was specified on the `go-table` component. The `name` column, however, is no longer `null` and will use the value passed to `wrapContent`.

The opposite scenario is also true:
```html
<go-table [wrapContent]="false">
  <go-table-column field="id"></go-table-column>
  <go-table-column field="name" [wrapContent]="true"></go-table-column>
</go-table>
```

See the logic of this [here](https://github.com/mobi/goponents/pull/820/files#diff-68fef6945ac75f361cff6412124644568c3fd5392e70a16bd82a9cee8b4f3892R123).

grahamhency:
@schneiderduketangoe I had to rebase and also explained my thought process, would you mind approving again?

projects/go-lib/src/lib/components/go-table/go-table-column.component.ts
schneiderduketangoe: Ah gotcha, so the idea is that null allows it to inherit from its parent element. Cool!

schneiderduketangoe (approved):
Good to go!

PR #824, opened 2021-08-16
[Chore] Panel Disable Option

### LINKED ISSUE #817: [Feature] Ability to disable a go-panel
I am requesting the ability to specify that a `go-panel` is disabled and not actionable. 

This should change the appearance of the panel to a disabled state and appear not-clickable. When a user clicks on the panel, nothing should happen. No event should be emitted.

jiaweicai92:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

Resolves #817 

## What is the new behavior?
Matches design provided by UX

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.html
+++ b/projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.html
@@ -1,7 +1,6 @@
 <!-- External Link -->
 <a
   class="go-panel"
-  [class.go-panel__disabled]="disablePanel"
   [target]="target ? target : '_blank'"
   [rel]="target ? target : 'noreferrer'"
   [ngClass]="panelClasses()"
--- a/projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.html
+++ b/projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.html
@@ -1,7 +1,7 @@
 <!-- External Link -->
 <a
   class="go-panel"
-  [disabled]="disablePanel" 
+  [class.go-panel__disabled]="disablePanel"
   [target]="target ? target : '_blank'"
   [rel]="target ? target : 'noreferrer'"
   [ngClass]="panelClasses()"
--- a/projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.html
+++ b/projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.html
@@ -1,6 +1,7 @@
 <!-- External Link -->
 <a
   class="go-panel"
+  [disabled]="disablePanel" 
   [target]="target ? target : '_blank'"
   [rel]="target ? target : 'noreferrer'"
   [ngClass]="panelClasses()"
@@ -17,6 +18,7 @@
 <ng-template #noLink>
   <button
     class="go-panel"
+    [disabled]="disablePanel"
     [ngClass]="panelClasses()"
     (click)="handleAction()">
     <span class="go-panel__title">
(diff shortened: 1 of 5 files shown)


projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.html
```diff
@@ -1,6 +1,7 @@
 <!-- External Link -->
 <a
   class="go-panel"
+  [disabled]="disablePanel"
```
grahamhency, line 4: So this won't work because `disabled` isn't a property on an `a` element. What do you think about adding a modifier class for the `go-panel` class instead?

```
.go-panel {
  &--disabled {
    opacity: 0.5;
  }
}

File: projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.scss [hunk 2]
```diff
@@ -116,4 +116,8 @@ button {
   &:focus {
     outline: none;
   }
+
+  &:disabled {
```
grahamhency, line 120: This will only apply to the `button` element and not the `a`

jiaweicai92: Thanks for pointing this out. Made some change,

projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.html
```diff
@@ -1,6 +1,7 @@
 <!-- External Link -->
 <a
   class="go-panel"
+  [class.go-panel__disabled]="disablePanel"
```
grahamhency, line 4: Should we use the `ngClass` on line 7 instead of this?

grahamhency (changes requested):

projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.html
jiaweicai92: Like your approach more. It is cleaner!

projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.scss
```diff
@@ -72,6 +72,11 @@ $base-medium: #767676 !default;
   &--header {
     color: $base-medium;
   }
+
+  &--disabled {
+    opacity: 0.5;
```
grahamhency, line 77: If you look at the action sheet, if we disable the first panel then its opacity is affected causing the element behind it to be shown and look weird. I'm not sure we can use opacity here.

An alternative would be to change the `color` property to use an `rgba` value with.

```scss
color: transparentize($theme-light-color, .5);

File: projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.scss [hunk 5]
```diff
@@ -116,4 +121,8 @@ button {
   &:focus {
     outline: none;
   }
+
+  &:disabled {
+    opacity: 0.5;
```
grahamhency, line 126: We can't use opacity here for the same reason.

grahamhency (changes requested):
I just looked at the view (which I should've done the first time around, my bad). I've got a comment on using opacity and also a concern around the hover state of the panel.

Screen Shot 2021-08-19 at 10 07 32 AM

If the panel is disabled, I don't think we should apply the hover state. What do you think?

projects/go-lib/src/lib/components/go-action-sheet/go-panel/go-panel.component.scss
```diff
@@ -2,6 +2,7 @@
 @import '../../../../../styles/mixins';
 
 $base-medium: #767676 !default;
+$theme-light-transparentized: $theme-light-color, .5;
```
grahamhency, line 5: Does this work? Seems like it should be:
```scss
$theme-light-transparentized: transparentize($theme-light-color, .5);

grahamhency (changes requested):

grahamhency (approved):
Took a look at the view, looks good to me. Approving! :+1:

PR #829, opened 2021-08-26
[Bug] Form Elements Errors

grahamhency:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)

Currently all of the Goponents form elements did not properly render the errors on the `FormControl`s they represented. We had logic there to handle it, but it was looking for a string array on the control rather than an object with key/value pairs as it should with Angular's `ValidationErrors` object on controls.

There is also currently quite a bit of shared logic amongst the form elements.

Some form elements currently do not have proper error states.

### Form Element Error Hints
This PR extracts the logic for rendering errors out of each form component and into its own. This means that each form component has been updated to render the new component. The new component contains the old logic (just in case its used somehow by someone), but also the new logic for looping over and reading the `ValidationErrors` on an Angular `FormControl`.

### Angular Validators Default Error Support
This PR adds default error messages for most of the Validators provided by Angular. These can be overridden, but will render automatically if one of the Angular Validators is added to a control.

### Abstract Typing
This adds a slightly more broad typing for the controls on each form component to allow the control to be an `AbstractControl` as well as a `FormControl`. The main reason for this is because doing a `form.get('controlName')` returns an `AbstractControl` rather than a `FormControl` and this can lead to some type mismatch errors.

### Form Base Component
This PR introduces the `GoFormBaseComponent`. This component does not get rendered anywhere, but get's extended off of by all of the other form components. This allows us to share common functionality amongst all the components and keep things more concise for each component.

### Form Service
This PR introduces the `GoFormService`. This service contains various methods that are used throughout the Goponents form components. This service may also be used externally as needed.

### Hint Design
This PR removes some old styling from the `go-hint` class. They're no longer italicized.

### Validation Errors Documentation
This PR added a card on the on the base [Forms Overview](https://tangoe.design/ui-kit/forms) page of the style guide to provide an example and better documentation on error handling by Goponents form components and how to interact with them.

- [ ] Yes
- [x] No

**This PR should not introduce any breaking changes, however, depending upon how the implementor has handled rendering form errors and their use of `ValidationErrors` could result in duplicate error messages on forms.** It is near impossible for us to know what affect it will have on those apps, but the corrected approach implemented within this PR should be the default and correct way of performing form validations henceforth.


jiaweicai92 (approved):
Looking good to me 🥇

PR #832, opened 2021-08-30
FEATURE: Added unordered list styles

### LINKED ISSUE #831: [Feature] Add styles for unordered lists
### Unordered Lists do not have styles
Found that there is no common class/styles for `<ul>` elements. Adding a class for this would allow common stylings of all ul moving forward. Main styles included are `list-style-type` and `list-style-position` along with some margin/line-height changes.

_**Example:**_
>.go-unordered-list
> We have a need for this class name for `<ul>` elements
> `<ul class="go-unordered-list"></ul>`

Adding a class into the _typography.scss file

swimtech20:
Added a class for unordered lists. Also added line-height to go-headings for text-wrapping

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Style Update (CSS)
- [x] Other... Please describe: New class for unordered lists

There are no common styles for `<ul>` elements

Resolves #831 

## What is the new behavior?
All `<ul>` moving forward will be styled the same way (using the goponent classname)

- [ ] Yes
- [x] No

### DIFF
--- a/projects/go-lib/styles/_typography.scss
+++ b/projects/go-lib/styles/_typography.scss
@@ -112,7 +112,7 @@ strong {
   list-style-position: outside;
   list-style-type: disc;
 
-  &__inner {
+  &--inner {
     list-style-type: circle;
   }
 }
--- a/projects/go-lib/styles/_typography.scss
+++ b/projects/go-lib/styles/_typography.scss
@@ -11,6 +11,7 @@ strong {
 .go-heading {
   font-family: $primary-font-stack;
   font-weight: $weight-regular;
+  line-height: 1.4;
   margin-bottom: $heading-margin-bottom;
 }
 
@@ -100,5 +101,16 @@ strong {
   &__item {
     line-height: 1.5;
     margin: $column-gutter--half 0;
+  }
+}
+
+.go-unordered-list {
+  line-height: 1.6;
+  list-style-position: outside;
+  list-style-type: disc;
+  margin: 0 0 2rem 1.1rem;
+
+  &__inner {
+    list-style-type: circle;
   }
 }
--- a/projects/go-lib/styles/_typography.scss
+++ b/projects/go-lib/styles/_typography.scss
@@ -11,7 +11,7 @@ strong {
 .go-heading {
   font-family: $primary-font-stack;
   font-weight: $weight-regular;
-  line-height: 1.4;
+  line-height: 1.3;
   margin-bottom: $heading-margin-bottom;
 }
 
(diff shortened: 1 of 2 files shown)


projects/go-lib/styles/_typography.scss
```diff
@@ -11,6 +11,7 @@ strong {
 .go-heading {
   font-family: $primary-font-stack;
   font-weight: $weight-regular;
+  line-height: 1.4;
```
grahamhency, line 14: So I looked through the Goponent's repo in an effort to be consistent and it looks like we do `1.3` in most places for line-height. (honestly we should probably have a variable for that). What do you think about changing this to `1.3`?

Additionally, this will affect the height of the heading elements through all the apps. What do you think about modifying the `$heading-margin-bottom` variable to say `1rem` to help accommodate for the increase in size? It will still be slightly larger, but I think that's fine and honestly looks better than the `1.5rem` it's at now.

projects/go-lib/styles/_typography.scss
```diff
@@ -102,3 +103,14 @@ strong {
     margin: $column-gutter--half 0;
   }
 }
+
+.go-unordered-list {
+  line-height: 1.6;
+  list-style-position: outside;
+  list-style-type: disc;
+  margin: 0 0 2rem 1.1rem;
```
grahamhency, line 111: Looking at the `go-ordered-list` class it seems like these should share some common styles. Thought's on trying to consolidate these a bit and make some of the styles match (like spacing, etc)?

grahamhency (changes requested):
Had a couple of comments. Also, would you mind writing some docs for these classes under the "typography" section? If you want to be extra cool you could write some docs for the ordered-list one as well. 😬

projects/go-lib/styles/_typography.scss
swimtech20: I like that idea

projects/go-lib/styles/_typography.scss
swimtech20: Good point!

swimtech20: Well, actually, they work a bit different from each other... @grahamhency

grahamhency: @swimtech20 do they? Aight. Just trying to make these as consistent as possible because they are structured the same in the DOM and once these classes are out there for people to use, if we end up changing the names or structure later that will be a breaking change! 🙃

swimtech20: @grahamhency can we maybe pair on this tomorrow? Maybe I'm misunderstanding something, but I do see your point...

swimtech20:
@grahamhency where is the typography section I need to write docs in? (mentioned above)

grahamhency (approved):

PR #1004, opened 2026-04-29
Angular upgradation from v14 to v19

jagadishchaudhari-tangoe:
Angular upgradation changes from V14 to V19.


projects/go-lib/tsconfig.lib.prod.json
```diff
@@ -1,6 +1,4 @@
 {
   "extends": "./tsconfig.lib.json",
-  "angularCompilerOptions": {
-    "enableIvy": false
-  }
+  "angularCompilerOptions": {}
 }
```
swimtech20: Needs a new line at the end of the file.

.vscode/tasks.json
```diff
@@ -0,0 +1,16 @@
...
+			"command": "npx",
+			"args": [
+				"ng",
+				"serve",
+				"--project=go-style-guide"
+			],
+			"isBackground": true
+		}
+	]
+}
```
swimtech20: Make sure new lines are at the end of all files.

projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.ts
```diff
@@ -17,12 +17,13 @@ import { GoConfigInterface } from '../../go-config.model';
...
-  animations: [
-    accordionAnimation
-  ]
+    selector: 'go-accordion-panel',
+    templateUrl: './go-accordion-panel.component.html',
+    styleUrls: ['./go-accordion-panel.component.scss'],
+    animations: [
+        accordionAnimation
+    ],
+    standalone: false
```
swimtech20: I'm worried that `standalone: false` will cause issues when Goponents is ingested because we will be making Mobile implement Standalone. Having components as standalone components will make data loading more efficient, simplify UTs, and on multiple levels improve maintainability. We will have to make every component standalone.

swimtech20 (changes requested):
Changes look good, but I have some concerns:

1. make sure all files end in a new line.
2. we want to utilize Standalone components as this will improve loading and maintainability. Default value is "true" so we just need to remove "standalone: false" entirely.
3. Publishing multiple upgraded versions of Angular at a time FORCES whoever ingests Goponents to upgrade to that version in order for their app to compile. It would be better practice to upgrade one Angular version at a time. I understand it might be too late for this PR, but please at least increment the goponent version so that we can get to 14, then to 19. And please publish a new version of Goponents for Angular 20, and another one for 21 moving forward.

@jiaweicai92 @jagadishchaudhari-tangoe

projects/go-lib/src/lib/components/go-accordion/go-accordion-panel.component.ts
jagadishchaudhari-tangoe: standalone: false has been removed and can be use it.

.vscode/tasks.json
jagadishchaudhari-tangoe: New lines are added to all files.

projects/go-lib/package.json
swimtech20: We should increment the Goponent version here to 1.18.4 so that we can ingest 1.18.3 while upgrading to Angular 14, then 1.18.4 when we continue to Angular 19.

swimtech20 (changes requested):

swimtech20:
What's the status on this? Are we able to resolve these conflicts and get this merged in? I know that Vijeta is working on upgrading Mobile and it'd be good to stay ahead...

projects/go-lib/package.json
swimtech20: Does the version get incremented during the publish process? Or do we manually need to specify?

.vscode/tasks.json
swimtech20: Looks like there is no new line here.

projects/go-lib/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "goponents",
-  "version": "1.18.3",
+  "version": "1.18.4",
```
swimtech20, line 3: Can we increment the version to a new major version? There are breaking changes in here and it might be nice to designate that with the version.

I'm thinking `2.0.0` or I guess I'd be okay with `1.19.0`. @jiaweicai92 thoughts?

swimtech20 (changes requested):
I noticed that there are still Module files in here.. the neat thing about Angular 17 is that each component is Standalone by default, and modules are no longer required. This means that you can import an individual components without having to import everything within the module (helps reduce compilation/run times).
 
However, considering that Goponents is essentially a node_module package and is ingested all together by other apps, maybe Modules are still needed here...?

Have you looked into `standalone` at all? Was there a decision made to intentionally keep all the modules?

.vscode/tasks.json
jagadishchaudhari-tangoe: New line is added.

projects/go-lib/package.json
jagadishchaudhari-tangoe: Yes, we plan to upgrade the version to 2.0.0 at release time, as this release includes breaking changes and substantial updates. A major version increment will better reflect the scope of the changes.

projects/go-lib/package.json
jagadishchaudhari-tangoe: Yes, we plan to upgrade the version to 2.0.0 at release time, as this release includes breaking changes and substantial updates. A major version increment will better reflect the scope of the changes.

jagadishchaudhari-tangoe (commented):
Thanks for the feedback and for raising the question around standalone components.

Yes, keeping the Module files after the Angular upgrade is intentional and currently serves an important purpose for the Goponents library.

While the components have been migrated to support the standalone architecture, we have retained the NgModule wrappers to maintain backward compatibility for existing consumer applications. Many applications currently consume Goponents through NgModule-based imports, and removing those module entry points at this stage would introduce a breaking API change and require coordinated updates across all consuming applications.

The current approach follows a dual-support strategy, where:

Components can leverage the standalone architecture introduced in newer Angular versions.
Existing consumers can continue using NgModule imports without disruption.

The primary objective of this effort was to upgrade Angular versions while preserving compatibility and minimizing migration risk for downstream applications. For that reason, we intentionally retained the module wrappers as compatibility facades.

That said, I agree that moving toward a standalone-first library architecture is the long-term direction. We can approach that gradually in a future initiative by defining a safe phased migration plan, including consumer adoption guidance, deprecation notices, and eventual retirement of module wrappers once all dependent applications have transitioned.

swimtech20 (changes requested):
After discussion with @jiaweicai92 , we came to the conclusion that we are not comfortable releasing a Goponent's update that jumps of more than one Angular version. I have warned about this before, and I think we got lucky releasing an update to upgrade Angular 9-14.

In this case, I know for a fact that Angular 17, 19, and 20 will cause major breaking changes to our downstream consumers. We need to close this PR and open a PR for each Angular version (that also increments Goponent version for each Angular upgrade). This will help Goponent consumers in the process of upgrading their own Angular versions (no version mismatch errors or compilation issues).

jagadishchaudhari-tangoe:
Closing this PR as per the updated release approach discussed with the development team.

Although this PR includes Angular upgrade changes from v15 to v19 along with related UI changes and fixes, it has been decided to first release only the Angular upgrade up to v14 with the required UI updates. The remaining version upgrades will be handled separately and released independently, one version at a time, to avoid potential ingestion issues for consuming applications.

Next steps:
- Verify the changes in the already created individual branches
- Perform unit testing on each branch
- Release the changes incrementally by increasing the minor version of goponent

Closing this PR for now as part of the phased rollout plan

PR #1011, opened 2026-06-10
TOP-5647 & TOP-5545 UI changes.

jagadishchaudhari-tangoe:
This PR consist the UI changes for the stories - TOP-5545 & TOP-5647

### DIFF
--- a/projects/go-lib/src/lib/components/go-input-group/go-input-group.component.ts
+++ b/projects/go-lib/src/lib/components/go-input-group/go-input-group.component.ts
@@ -1,7 +1,6 @@
-﻿import { Component, Input, ViewEncapsulation } from '@angular/core';
+﻿import { Component, Input } from '@angular/core';
 
 @Component({
-  encapsulation: ViewEncapsulation.None,
   selector: 'go-input-group',
   templateUrl: './go-input-group.component.html',
   styleUrls: ['./go-input-group.component.scss']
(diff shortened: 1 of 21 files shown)


projects/go-lib/src/lib/components/go-input-group/go-input-group.component.ts
```diff
@@ -0,0 +1,13 @@
+﻿import { Component, Input, ViewEncapsulation } from '@angular/core';
+
+@Component({
+  encapsulation: ViewEncapsulation.None,
```
swimtech20: Having `ViewEncapsulation.None` set will apply all component styles globally. In this case, `:host` will be display block with a margin bottom everywhere, not just for this `go-input-group.component`. Just double checking this is what is wanted/needed...?

swimtech20 (changes requested):

projects/go-lib/src/lib/components/go-input-group/go-input-group.component.ts
jagadishchaudhari-tangoe: Thanks for pointing this out. I have reviewed the concern and removed the ViewEncapsulation.None usage. The component styles are now scoped appropriately, and the global styling impact is no longer present.

swimtech20 (approved):

PR #1040, opened 2026-07-21
chore: upgrade Angular framework from v19 to v20

jagadishchaudhari-tangoe:
This PR upgrades the application from Angular v19 to Angular v20 and applies the recommended framework migrations. The goal is to keep the application aligned with the latest Angular release, benefit from performance improvements, security updates, and maintain long-term support.


jiaweicai92 (approved):

swimtech20 (approved):
@jagadishchaudhari-tangoe something to keep in mind: Angular 20 will allow current *ngIf, *ngFor, etc directives for now, but they are deprecated and will soon cause errors in future Angular versions. What I advised Vijeta to do in Mobile upgrade is to skip them for now and just get to Angular 21, and _then_ go back and fix all the HTML templates to reflect this deprecation.

Not sure if you can see it in your IDE, but this will need to be updated soon. It's fine for now since we are trying to get to ng v21, but it will fail compilation at some point.

